PR #5, opened 2022-02-10
Stl 1821 erica api v.2

VictorDelCampo:
# Short Description
- Added v2 version leaned to FastAPI structure (v1 adapted to that structure).

# Changes
- Version 01 restructured, leaned to FastAPI structure
- Create new version 02 with only "tax_offices" implemented
- Other requests will communicate with the job queue and therefore for now NotImplementedError, but with an exception catch for producing a dummy response in accordance with the specification
- There are 2 preconfigured docs urls that can be accessed as soon as the erica service is started:
-- Swagger UI served at /docs
-- ReDoc UI served at /redoc

# Feedback
- I am not really happy with the big class/variable names, specially for the unlock_code requests.
- But i wanted to maintain all params/variables passed in the v1 requests
- The success responses should be defined when the actual implementation is done, right?
- About the documentation, are the docs urls what was meant in the ticket?
- Grundsteuer folder in v2 is empty. What about informing the team? How should it be done and what is to say?


VictorDelCampo:
@JulianRoesner 
- Variable "include_elster_response" removed from all payload models.
- Variable "ttlInMinutes" unchanged, using camel case would means the actual name of the variable in the request payload would differ from your Tech Spec
- Response models for v2 were added but for the get_job-requests the default 422 ValidationError from openAPI will be displayed in the docs. Actually it is not possible that this error occurs since if we don't pass a request_id then we would be trying to access the API of the POST request and we would get a HTTp 405 Method not allowed. There is a way to completely removed the default 422 from openAPI but this would mean inherting from FastAPI and rewriting the method openapi (https://github.com/tiangolo/fastapi/issues/3424)
- Creating response models for v1 would mean updating all API resources so that we return JSONRespones and not HTTPException, but you said, v2 would be important and v1 only if possible and easy?

If you find all changes OK, just approve and I would merge to main and let Frede now about our changes.

erica/api/v2/endpoints/rente/unlock_code.py
```diff
@@ -0,0 +1,93 @@
...
+
+from erica.pyeric.utils import generate_dummy_error_response
+from erica.request_processing.erica_input import UnlockCodeRequestDataWithTtl, \
+    UnlockCodeActivationDataWithTtl, UnlockCodeRevocationDataWithTtl, ErrorRequestQueue, ResponseGetFromQueue
+
+router = APIRouter()
+
+
+@router.post('/requests', status_code=status.HTTP_201_CREATED, responses={422: {"model": ErrorRequestQueue}})
+def request_unlock_code(unlock_code_request_ttl: UnlockCodeRequestDataWithTtl):
```
JulianRoesner, line 14: If we change the name of the endpoint to fsc/request (s. above) we probably should do some renaming here, too.

JulianRoesner, line 14: If we change the name of the endpoint to fsc/request (s. above) we probably should do some renaming here, too.

erica/api/v2/responses/model.py
```diff
@@ -0,0 +1,9 @@
+from erica.request_processing.erica_input import ErrorRequestQueue, ResponseGetFromQueue
+
+response_model_post_to_queue = {
+    201: {"description": "Job was successfully submitted to the queue and the job id was returned."},
+    422: {"model": ErrorRequestQueue, "description": "Job could not be submitted to the queue."}}
+
+response_model_get_from_queue = {
+    200: {"model": ResponseGetFromQueue, "description": "Job status was successfully retrieved from the queue."},
+    500: {"model": ErrorRequestQueue, "description": "Job status could not be retrieved from the queue."}}
```
JulianRoesner, line 9: How is this used?

erica/request_processing/erica_input.py
```diff
@@ -237,3 +237,48 @@ class UnlockCodeRevocationData(BaseModel):
...
+class ErrorRequestQueue(BaseModel):
+    errorCode: str
+    errorMessage: str
+
+
+class ResponseGetFromQueue(BaseModel):
+    processStatus: str
+    payload: Optional[str] = None
+    errorCode: Optional[str] = None
+    errorMessage: Optional[str] = None
```
JulianRoesner, line 282: I am wondering if it might make sense to put them into a different file/folder_structure?

JulianRoesner (commented):
Very nice. I just have a question about the response_models and I am wondering if there would be a better way to structure the erica_input file

erica/api/v2/api_v2.py
```diff
@@ -0,0 +1,16 @@
...
+from erica.api.v2.endpoints.rente import unlock_code, est, tax
+
+
+api_router_02 = APIRouter()
+api_router_02.prefix = '/02'
+
+api_router_02.include_router(ping.router, prefix="/ping", tags=["Ping"])
+api_router_02.include_router(est.router, tags=["Steuererklärung"])
+api_router_02.include_router(tax.router, tags=["Finanzverwaltung"])
+api_router_02.include_router(unlock_code.router, prefix="/unlock_code", tags=["Freischaltcode"])
```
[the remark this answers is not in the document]
VictorDelCampo: We could do that but i thought we were using English for the routes. Fsc would be anyway just a short form but a short form in German

erica/api/v2/responses/model.py
VictorDelCampo: Those responses are used in every post/get route for the job queue and define how the returned structure for the specified HTTP code should look like. This is necessary for the docs to display them. They are addressed with the "responses" attribute of the route

erica/__init__.py
```diff
@@ -4,6 +4,7 @@
 from prometheus_client import Gauge
 from prometheus_fastapi_instrumentator import Instrumentator
 
+from erica.api.api import api_router
 from erica.pyeric.eric import verify_using_stick
 
 app = FastAPI()
```
punknoir101, line 10: Why we don't use the standard way to versioning APIs? For Fast API there is [fast api versioning package](https://github.com/DeanWay/fastapi-versioning) which uses annotations (The more common way, I would say.)

here:
`
app = FastAPI(title="Erica Api")
app = VersionedFastAPI(app, default_api_version=(1))
`

and annotation like this:
`
@version(1, 2)
`

in routes

erica/api/v2/api_v2.py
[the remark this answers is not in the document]
JulianRoesner: >  i thought we were using English for the routes

Yes that was the reason we used unlock_code here. The reason it came up is that this term seems to confuse sometimes. We decided that we will use English in the code wherever it makes sense but use the German terms if they refer to a special concept. Therefore, both would work here

erica/api/v2/responses/model.py
JulianRoesner: Ok, I just do not understand how they are referenced

erica/request_processing/erica_input.py
JulianRoesner: Exactly

erica/__init__.py
JulianRoesner: > Why we don't use the standard way to versioning APIs? 

Can you elaborate what you mean with the standard way?

I like the fast api versioning package. However, we could only use it for version 2 onwards as this would change the routes

erica/api/v2/api_v2.py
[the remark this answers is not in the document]
VictorDelCampo: Okidok! I will change it to fsc

erica/api/v2/endpoints/rente/unlock_code.py
VictorDelCampo: Will be changed to fsc/request and the name of the method would be: request_fsc, I will also change the variables

VictorDelCampo: Will be changed to fsc/request and the name of the method would be: request_fsc, I will also change the variables

erica/request_processing/erica_input.py
VictorDelCampo: Will be done :)

erica/api/v2/responses/model.py
VictorDelCampo: With the responses attribute of the route:
`@router.post('/est_validations', status_code=status.HTTP_201_CREATED, responses=response_model_post_to_queue)`

erica/__init__.py
VictorDelCampo: I'v just made an implementation of that library in our erica and looks quite nice! We can adapt how the versions are formatted, including a prefix, like this:

`app = VersionedFastAPI(app,
                       version_format='{major}',
                       prefix_format='/0{major}')`

And then we only use the major version decorator:
`@version(2)`

This way we can also update the version markers for 01

My concerns on this library (following our specification for adding a new external library):
- last release was 1 year ago
- last commit 6 months ago
- 20 open issues (5 of them from 2020 and rest from 2021 but 2 from 2022)
- 4 open PR (all of them from 2021)

I have just pushed my implementation with that library. One BIG advantage of using the library is that we have the docs in Swagger in the corresponding version urls:
/01/docs
/02/docs

erica/api/v2/responses/model.py
JulianRoesner: Somehow, I overlooked that. Thank you 🙈

erica/__init__.py
JulianRoesner: >This way we can also update the version markers for 01

Nice.

>My concerns on this library(following our specification for adding a new external library):
last release was 1 year ago
last commit 6 months ago
20 open issues (5 of them from 2020 and rest from 2021 but 2 from 2022)
4 open PR (all of them from 2021)

Thanks for checking. Yes I think I agree with your concerns. Apparently, it started more as a proof of concept. Have [a look at this article](https://medium.com/geoblinktech/fastapi-with-api-versioning-for-data-applications-2b178b0f843f). It mentions that there might be problems with custom error handling. Maybe we should have a look into that.
Also they have [an issue open asking if the project is still active](https://github.com/DeanWay/fastapi-versioning/issues/65).

VictorDelCampo: The only advantage of using this library would be, for me, having by default different docs urls for each version. FastAPI [offers](https://fastapi.tiangolo.com/tutorial/metadata/) the possibility to change both urls (Swagger and Redoc) but we would need two different FastAPI instances in which we could separately add the routers of each version and then setting specific docs url, like stated [here](https://github.com/tiangolo/fastapi/issues/381#issuecomment-584132553). But the question would be, do we want separate docs urls?

Using this library, we still need to decorate every API method we define or we have defined. This was't necessary for the prefix versioning way.

Regarding the problems with custom error handling, it would be a stopper for us if we were already using those, but as far as I know, we don't. Nevertheless the project has on github many open issues and most of them differ from each other. In addition the unanswered question of the project being active or not for almost a month now.

In my opinion I would avoid using this library. It looks inactive since October 2021:
image

WDYT?

punknoir101: > > Why we don't use the standard way to versioning APIs?
> 
> Can you elaborate what you mean with the standard way?
> 
> I like the fast api versioning package. However, we could only use it for version 2 onwards as this would change the routes

Standard way: [open api standard](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#versions) / [open api standard](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#server-object)

punknoir101: > The only advantage of using this library would be, for me, having by default different docs urls for each version. FastAPI [offers](https://fastapi.tiangolo.com/tutorial/metadata/) the possibility to change both urls (Swagger and Redoc) but we would need two different FastAPI instances in which we could separately add the routers of each version and then setting specific docs url, like stated [here](https://github.com/tiangolo/fastapi/issues/381#issuecomment-584132553). But the question would be, do we want separate docs urls?
> 
> Using this library, we still need to decorate every API method we define or we have defined. This was't necessary for the prefix versioning way.
> 
> Regarding the problems with custom error handling, it would be a stopper for us if we were already using those, but as far as I know, we don't. Nevertheless the project has on github many open issues and most of them differ from each other. In addition the unanswered question of the project being active or not for almost a month now.
> 
> In my opinion I would avoid using this library. It looks inactive since October 2021: image
> 
> WDYT?

Why do you think packages have to be under heavy development?

JulianRoesner: Yes, I agree with you. I think that using the decorators actually would have an advantage as we would see what is outdated quite easily.

However, imo the library does not look active enough that we should use it for such a small advantage where it does not give us that much to maintain. And we will probably most of the time only have on active version running anyways

VictorDelCampo: We can already see what is outdated fast by checking the folder of the api, since we moved 01 to folder v1 and the new 02 to folder v2.

I would then revert the changes I made when introducing FastAPI-versioning and go on with the prefix technique within the API router?

JulianRoesner: > Standard way: [open api standard](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#versions) / [open api standard](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#server-object) / [article](https://www.akana.com/blog/api-versioning)

Good, I was not sure if that is what you meant. I agree with you. There is no reason to not use the standard way for v2 onwards. Unfortunately, not for v1.

> Why do you think packages have to be under heavy development?

In general, they do not need to be under heavy development, but if we rely on them and we can not be sure that they are updated regularly, this is a dependency that might lead to problems (e.g. we still are not able to use Python 3.10 because of some dependencies). Furthermore, as this library is basically a proof-of-concept and already shows some problems (the mentioned problem with custom error handling) I would assume that we should not rely on that, unless it reduces a lot of workload. I do not see that but maybe I am overlooking something?

JulianRoesner: > We can already see what is outdated fast by checking the folder of the api, since we moved 01 to folder v1 and the new 02 to folder v2.

Yes, I thought so.

> I would then revert the changes I made when introducing FastAPI-versioning and go on with the prefix technique within the API router?

That's what I would say. However, we probably should rename the version two to follow the standard @punknoir101 mentioned?

VictorDelCampo: Meaning 2.0 and not 02?

punknoir101: It's about the open api standard and the api documentation - if you do it "your" way it will work, but it's not properly documented - the swagger.json/open-api.json file would be wrong then. So other services that will follow open api standard would have problems to consume different versions. **Open Api json is the contract and should be properly documented!**

VictorDelCampo: Not really getting your point. The open api standard has a field within called "openapi", which is a required string and " MUST be the [semantic version number](https://semver.org/spec/v2.0.0.html) of the [OpenAPI Specification version](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#versions) that the OpenAPI document uses."

And it follows "This is not related to the API [info.version](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#infoVersion) string."

That field "version" within the object "info" of the specification is just a simple string. And this value can be set through the FastAPI instance. If we don't use the FastAPI-Versioning we could just create 1 FastAPI instance for each API version we have and defined the version in the format we want.

WDYT?

erica/api/v2/endpoints/rente/fsc.py
```diff
@@ -0,0 +1,102 @@
...
+    :param fsc_revocation_ttl: payload with TTL and the JSON input data for the revocation.
+    """
+    try:
+        raise NotImplementedError()
+    except NotImplementedError:
+        logging.getLogger().info("Could not revoke unlock code", exc_info=True)
+        return JSONResponse(status_code=422, content=generate_dummy_error_response())
+
+
+@router.get('/revocation/{request_id}', status_code=status.HTTP_200_OK, response_model=ResponseGetFromQueue,
```
JulianRoesner, line 89: i was just wondering if we should use special response_models for the different responses that contains the actual data that will be send and not only the payload field. Wdyt?

VictorDelCampo: Yes that would be better! So I will refer to the response payloads of v1 for this, thanks!

erica/api/v2/responses/model.py
```diff
@@ -1,9 +1,40 @@
...
+
+response_model_get_unlock_code_activation_from_queue = {
+    200: {"model": ResponseGetUnlockCodeRequestAndActivationFromQueue,
+          "description": "Job status of an unlock code activation was successfully retrieved from the queue."},
+    500: model_error_request_queue}
+
+response_model_get_unlock_code_revocation_from_queue = {
+    200: {"model": ResponseGetUnlockCodeRevocationFromQueue,
+          "description": "Job status of an unlock code revocation was successfully retrieved from the queue."},
+    500: model_error_request_queue}
```
JulianRoesner, line 40: 👍

JulianRoesner (approved):
LGTM! We should have a check that everything works on staging as part of QA.

VictorDelCampo:
@JulianRoesner  I added an enum for the "Status" and modified the base class of the response model when getting a job from the queue

erica/__init__.py
JulianRoesner: Sorry to open that up again. While looking for resources for another discussion I came upon [this article ](https://www.freecodecamp.org/news/how-to-version-a-rest-api/) and [this one](https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#restful) that are both talking about including the api version in the uri and using v1 or v2 for that.

The second article was also the origin of the v1 in our code base. So, as far as I see it, this would be a standard way to expose our version through the url and then we could still version the whole api through FastApi's spec.

@punknoir101 @VictorDelCampo

erica/request_processing/erica_input/v2/erica_input.py
```diff
@@ -41,10 +40,18 @@ class ErrorRequestQueue(BaseModel):
...
 
 
+class Status(Enum):
+    PENDING = "Pending"
+    PROCESSING = "Processing"
+    FAILED = "Failed"
+    COMPLETED = "COMPLETED"
+
+
 class ResponseGetFromQueue(BaseModel):
```
JulianRoesner, line 50: Shouldn't we then somehow specify that we will also use this model for successful cases?

erica/request_processing/erica_input/v2/erica_input.py
```diff
@@ -41,10 +40,18 @@ class ErrorRequestQueue(BaseModel):
     errorMessage: str
 
 
+class Status(Enum):
+    PENDING = "Pending"
+    PROCESSING = "Processing"
+    FAILED = "Failed"
+    COMPLETED = "COMPLETED"
```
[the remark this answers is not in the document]
JulianRoesner: Nice idea with the enum 👍

erica/request_processing/erica_input/v2/erica_input.py
```diff
@@ -41,10 +40,18 @@ class ErrorRequestQueue(BaseModel):
     errorMessage: str
 
 
+class Status(Enum):
+    PENDING = "Pending"
```
JulianRoesner, line 44: We do not need to show to the outside that a job is pending imo. It should be enough to set it to processing

JulianRoesner (approved):
Mostly some stuff that I did not specify correctly. If you do not agree, let's talk.

Great idea with the enum

erica/request_processing/erica_input/v2/erica_input.py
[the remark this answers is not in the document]
VictorDelCampo: I copied the possible values of the enum from the tech spec:
image
But yes, "COMPLETED" should not be in capital lettres

erica/request_processing/erica_input/v2/erica_input.py
VictorDelCampo: So we don't need the status "Pending"? just processing, failed and completed?

erica/request_processing/erica_input/v2/erica_input.py
VictorDelCampo: I will name all response models for the HTTP 200 like "SuccessResponse..."

erica/request_processing/erica_input/v2/erica_input.py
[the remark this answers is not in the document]
VictorDelCampo: In accordance to the tech spec, we will only have 3 states: Processing, Failure and Success

PR #10, opened 2022-02-21
Add additional test idnr for production environment

JulianRoesner:
# Short Description
We sometimes have the problem that we want to test more than once per day the flow of request, activate and cancel unlock_code on production. We only have one test_idnr that we can use for that and are limited to performing this only once per day.
Therefore, this adds one more test IdNr to be used in production

# Changes
- Use an array instead of a string
- Adapt tests

# Feedback
- Do you see any problem with this?


frederike-ramin (approved):
Looks great. 👍 
Also the choice of the IdNr makes a lot of sense to me. :)

PR #17, opened 2022-03-11
Integrate new structure

JulianRoesner:
# Short Description
This adds the new folder structure from main-develop to the erica repository. Furthermore, I also added integration tests that we can use to test the application from the outside (I, for example, tested the docker image with those tests). I would propose that we add a step to our cicd-pipeline, where we run those tests on a pod accessing the deployed erica. That way we can ensure that changes do not have any bad effects on our integrated solution. 
We can exchange those in the future maybe through some kind of contract based testing.

# Changes
- Restructure the repository
- Add tests
- Fix a small error where we did accept incorrect state_abbreviations and would get a key error because of that.

# Feedback
- Any?


JulianRoesner:
The security hotspots can be resolved imo because they only affect some strings that are put into the xml not real http calls and I do not see a problem with the logger configuration

punknoir101 (commented):
OK, just one small thing: [tests/apllication/.gitignore](https://github.com/digitalservice4germany/erica/pull/17/files#diff-5face74ffcc8a549dcd3e6c11ed02adf9039e0fb13f57ba81af48ed47d3d2348) --> folder name with wrong spelling tests/a...

punknoir101 (approved):

PR #19, opened 2022-03-14
feat: introduce  job queue within clean architecture

punknoir101:
# Short Description
- introduce clean architecture
- Durchstich for FSC

# Changes
-  base configuration, models, repositories, services, DTOs for job queue

# Feedback
- Which parts of the code would you like feedback on?
- Questions that are open are also good (e. g. "Do you see any tests missing for class x?")


docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    restart: always
+    expose:
+      - 6379
+    ports:
+      - "6379:6379"
+  db:
+    image: postgres:latest
+    restart: always
+    environment:
+      - POSTGRES_USER=postgres
```
SannyNguyenHung, line 14: if this the local docker compose then I'm totally fine with this but we should also add a docker-compose.ci.yml for the secrets

SannyNguyenHung (approved):
It's okay for me now, didn't see any crucial impacts for the legacy code so far.

JulianRoesner (commented):
I was not able to go through every single file independently, yet. I have added a small change to ensure that building the docker image still works. I have tested the image with the integration tests. Therefore, I would assume that there is nothing fundamentally breaking through the new code as it is mostly dead code. Nevertheless, we will need to make sure to clean it up and test it a lot with future PRs.
And: We do not have to discuss this now, but personally, I am not such a big fan of a separate lib folder outside of the erica_module.

punknoir101:
> > Why do we need to add those` __init__.py` files?
> 
> Older versions of Python needed a `__init__.py` file to tell Python that a folder is a module. Newer Python versions do not need the **init**.py files. However, Pycharm adds this automatically.
> 
> I would propose we clean that up and delete those `__init__.py` files

:white_check_mark:

PR #21, opened 2022-03-16
Gstl gebaeude

shrzaf:
# Short Description
Extends Grundsteuer XML generation to support Gebäude.


erica/elster_xml/grundsteuer/elster_data_representation.py
```diff
@@ -3,15 +3,146 @@
...
+        if wohn_unter60.E7403131 > 0:
+            self.Wohn_Unter60 = wohn_unter60
+
+        wohn_60bis100 = EWohn60bis100(flaechen)
+        if wohn_60bis100.E7403141 > 0:
+            self.Wohn_60bis100 = wohn_60bis100
+
+        wohn_ab100 = EWohnAb100(flaechen)
+        if wohn_ab100.E7403151 > 0:
+            self.Wohn_ab100 = wohn_ab100
```
frederike-ramin, line 97: This is I think a bit the other way around than how I pictured the process. So far, the XML nodes just got the information they truly needed. This way, we also won't have to know the specific E-numbers of another class here (e.g.E7403131). So I'd rather construct three lists (unter60 etc) and then give these lists to the sub-nodes.
(and probably also make them english. :x

tests/request_processing/erica_input/v2/test_grundsteuer_input.py
```diff
@@ -1,8 +1,174 @@
 import pytest
```
frederike-ramin, line 1: Can you also split the test file if you split the original input file?

tests/request_processing/erica_input/v2/test_grundsteuer_input.py
```diff
@@ -1,8 +1,174 @@
...
+    def test_on_camel_case_should_map_to_snake_case(self):
+        input_data = {"isAb1949": True}
+
+        result = Ab1949.parse_obj(input_data)
+
+        assert result.is_ab1949 is True
+
+    def test_foo(self):
+        gebaeude = SampleGebaeude().with_baujahr("1950").build()
+        print(gebaeude)
```
frederike-ramin, line 21: ? 😉

tests/samples/grundsteuer_sample_data.py
```diff
@@ -2,6 +2,74 @@
 import datetime
 
 from erica.request_processing.erica_input.v2.grundsteuer_input import GrundsteuerData
+from erica.request_processing.erica_input.v2.grundsteuer_input_gebaeude import Gebaeude
+
+
+class SampleGebaeude:
```
frederike-ramin, line 8: I really really like it. Could by any chance also to this for eigentuemer? 🙈

erica/request_processing/erica_input/v2/grundsteuer_input_eigentuemer.py
```diff
@@ -0,0 +1,85 @@
...
+
+class Verheiratet(PossiblyAliasedInput):
+    are_verheiratet: bool
+
+
+class Person(PossiblyAliasedInput):
+    persoenlicheAngaben: PersoenlicheAngaben
+    adresse: Adresse
+    telefonnummer: Optional[Telefonnummer]
+    steuer_id: Optional[SteuerId]
```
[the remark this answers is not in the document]
frederike-ramin: Oh! I see, you just included it as a generator! Genius!

erica/request_processing/erica_input/v2/grundsteuer_input_gebaeude.py
```diff
@@ -0,0 +1,114 @@
...
+    weitere_wohnraeume_details: Optional[WeitereWohnraeumeDetails]
+    garagen: Garagen
+    garagen_anzahl: Optional[GaragenAnzahl]
+
+    @root_validator
+    def baujahr_must_be_present_if_ab_1949(cls, values):
+        v = values.get("baujahr")
+        if values.get("ab1949").is_ab1949 is True and (not v or not v.baujahr):
+            raise ValueError('has to be set if is_ab1949')
+        return values
```
frederike-ramin, line 77: Why did you decide on a root validator instead of 

```
@validator("baujahr", always=True)
    def baujahr_must_be_present_if_ab_1949(cls, v, values):
        if values.get("ab1949").is_ab1949 is True and (not v or not v.baujahr):
            raise ValueError('has to be set if is_ab1949')
        return v
```
This way I think it's clearer which field is validated?

erica/request_processing/erica_input/v2/grundsteuer_input_gebaeude.py
```diff
@@ -0,0 +1,114 @@
...
+    wohnflaechen: Optional[Wohnflaechen]
+    weitere_wohnraeume: WeitereWohnraeume
+    weitere_wohnraeume_details: Optional[WeitereWohnraeumeDetails]
+    garagen: Garagen
+    garagen_anzahl: Optional[GaragenAnzahl]
+
+    @root_validator
+    def baujahr_must_be_present_if_ab_1949(cls, values):
+        v = values.get("baujahr")
+        if values.get("ab1949").is_ab1949 is True and (not v or not v.baujahr):
```
frederike-ramin, line 75: Instead of checking `not v.baujahr` here, we could just remove the Optional type on the field? So this way, gebaeude.baujahr can never be set without gebaeude.baujahr.baujahr being set?

erica/request_processing/erica_input/v2/grundsteuer_input_gebaeude.py
```diff
@@ -0,0 +1,114 @@
...
+            raise ValueError('has to be set if is_ab1949')
+        return values
+
+    @root_validator
+    def kernsanierungsjahr_must_be_present_if_kernsaniert(cls, values):
+        v = values.get("kernsanierungsjahr")
+        if values.get('kernsaniert').is_kernsaniert is True and (not v or not v.kernsanierungsjahr):
+            raise ValueError('has to be set if is_kernsaniert')
+        return values
```
frederike-ramin, line 84: The two comments above go for all root validators, I think.
Also: Nice that you thought about including them. Precisely the validations I'd have included, too. 👍

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+
+
+class TestEWeitereWohn:
+    def test_should_construct_from_dict_correctly(self):
+        weitere_wohnraeume = WeitereWohnraeumeDetails.parse_obj({"anzahl": 2, "flaeche": 42})
+
+        result = EWeitereWohn(weitere_wohnraeume)
+
+        assert result.E7403121 == 2
+        assert result.E7403122 == 42
```
frederike-ramin, line 137: You dodn't want to include the len(vars()) test?

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+        assert result.E7403171 == 2
+
+
+class TestEAngDurchschn:
+    def test_wohnflaeche_under_60_should_set_others_to_none(self):
+        gebaeude = SampleGebaeude().with_wohnflaeche(59).parse()
+
+        result = EAngDurchschn(gebaeude)
+
+        assert result.Wohn_Unter60 is not None
```
frederike-ramin, line 155: Should we maybe instead test then `assert result.Wohn_Unter60 == EWohnUnter60(gebaeude.wohnflaeche)` or similar?

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+
+        assert result.E7403114 == "1959"
+
+    def test_on_vor_1949_should_not_contain_baujahr(self):
+        gebaeude = SampleGebaeude().with_wohnflaeche(42).with_baujahr("1959").parse()
+        gebaeude.ab1949.is_ab1949 = False
+
+        result = EAngWohn(gebaeude)
+
+        assert result.E7403114 is None
```
frederike-ramin, line 207: How about testing the vor1949 flag?

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+
+    def test_on_kenrsaniert_should_contain_kernsanierungsjahr(self):
```
frederike-ramin, line 209: Typo! :P

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+        gebaeude.weitere_wohnraeume.has_weitere_wohnraeume = False
+
+        result = EAngDurchschn(gebaeude)
+
+        assert result.Weitere_Wohn is None
+
+
+class TestEAngWohn:
+    def test_on_ab_1949_should_contain_baujahr(self):
+        gebaeude = SampleGebaeude().with_wohnflaeche(42).with_baujahr("1959").parse()
```
frederike-ramin, line 195: Why construct withWohnflaeche() in the following tests?

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+        assert result.E7403116 is None
+
+    def test_on_garagen_should_contain_set_anzahl(self):
+        gebaeude = SampleGebaeude().with_wohnflaeche(42).with_garagen(2).parse()
+
+        result = EAngWohn(gebaeude)
+
+        assert result.Garagen.E7403171 == 2
+
+    def test_on_no_garagen_should_contain_not_contain_garagen(self):
```
frederike-ramin, line 246: Typo?

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+        gebaeude = SampleGebaeude().with_wohnflaeche(42).with_garagen(2).parse()
+        gebaeude.garagen.has_garagen = False
```
frederike-ramin, line 248: Why do you first set garagen just to then unset them?

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+        result = EAngWohn(gebaeude)
+
+        assert result.Garagen is None
+
+    def test_should_set_ang_durchschn(self):
+        gebaeude = SampleGebaeude().with_wohnflaeche(42).parse()
+
+        result = EAngWohn(gebaeude)
+
+        assert result.Ang_Durchschn is not None
```
frederike-ramin, line 259: You know hwat I wanna say. :D

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -351,7 +595,7 @@ def test_attributes_set_correctly(self):
         assert result.GW1 == EGW1(grundsteuer_obj.eigentuemer)
         assert result.xml_attr_version == "2"
         assert result.xml_attr_xmlns == "http://finkonsens.de/elster/elstererklaerung/grundsteuerwert/e88/v2"
-        assert len(vars(result)) == 4
+        assert len(vars(result)) == 5
```
frederike-ramin, line 598: Can you then also test for the new GW2 exactly as for GW1?

tests/request_processing/erica_input/v2/test_grundsteuer_input.py
```diff
@@ -1,8 +1,174 @@
...
+class TestWeitereWohnraeume:
+    def test_on_camel_case_should_map_to_snake_case(self):
+        input_data = {"has_weitere_wohnraeume": True}
+
+        result = WeitereWohnraeume.parse_obj(input_data)
+
+        assert result.has_weitere_wohnraeume is True
+
+
+class TestWeitereWohnraeumeFlaeche:
```
frederike-ramin, line 69: Do you mean WeitereWohnraeumeDetails?

tests/request_processing/erica_input/v2/test_grundsteuer_input.py
```diff
@@ -1,8 +1,174 @@
...
+    def test_on_camel_case_should_map_to_snake_case(self):
+        input_data = {"anzahl": "2", "flaeche": "42"}
+
+        result = WeitereWohnraeumeDetails.parse_obj(input_data)
+
+        assert result.anzahl == 2
+        assert result.flaeche == 42
```
frederike-ramin, line 76: Also: There isn't really a camelcase here?

tests/request_processing/erica_input/v2/test_grundsteuer_input.py
```diff
@@ -1,8 +1,174 @@
...
+
+
+class TestGebaeude:
+    def test_on_is_ab1949_but_no_baujahr_should_raise_error(self):
+        gebaeude = SampleGebaeude().with_wohnflaeche("42").with_baujahr().build()
+
+        with pytest.raises(ValidationError):
+            Gebaeude.parse_obj(gebaeude)
```
frederike-ramin, line 84: As you explicitly included it: Should we also test if gebaeude.baujahr.baujahr is None but gebaeude.baujahr is set?

erica/request_processing/erica_input/v2/possibly_aliased_input.py
```diff
@@ -0,0 +1,8 @@
+from humps import camelize
+from pydantic import BaseModel
+
+
+class PossiblyAliasedInput(BaseModel):
```
frederike-ramin, line 5: I think we should rename this, because it is now not aliased anymore, but camelizable or something along the lines of it^^
Also: We might not even need the ` allow_population_by_field_name` option anymore.

frederike-ramin (commented):
Wow, nice!

Just a few smaller comments and some possible discussion points. Especially:

-File Structure: I feel like the `erica_input/v2/` file structure should be the same as the `elster_xml` file structure, it's probably easier to understand that way?
- Naming of the tests

Great job with the camel-casing, too!
And note to self: I actually *really* like the new structure - reviewing this PR was actually fun!

erica/elster_xml/grundsteuer/elster_data_representation.py
```diff
@@ -3,15 +3,146 @@
...
+@dataclass
+class EWohn60bis100:
+    E7403141: int = 0
+    E7403142: int = 0
+
+    def __init__(self, flaechen: List[int]):
+        for flaeche in flaechen:
+            if 60 <= flaeche < 100:
+                self.E7403141 += 1
+                self.E7403142 += flaeche
```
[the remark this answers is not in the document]
shrzaf: It was for my own reference, and I thought it would be helpful to leave them there once I was done, particularly the first instance of the classes with the pattern anzahl-flaeche and the fields in EAngWoh, since they can continue to serve as reference in case we need to make changes (or even read the code) here later, and also they're really minimal.

erica/elster_xml/grundsteuer/elster_data_representation.py
```diff
@@ -3,15 +3,146 @@
...
+    Wohn_ab100: Optional[EWohnAb100] = None
+    Weitere_Wohn: Optional[EWeitereWohn] = None
+
+    def __init__(self, input_data: GebaeudeInput):
+        flaechen: List[int] = []
+        if input_data.wohnflaeche:
+            flaechen.append(input_data.wohnflaeche.wohnflaeche)
+        elif input_data.wohnflaechen:
+            flaechen.append(input_data.wohnflaechen.wohnflaeche1)
+            flaechen.append(input_data.wohnflaechen.wohnflaeche2)
```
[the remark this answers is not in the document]
shrzaf: True, and this is not the only case where we can optimize erica input. However, in order for this to work, we would need to add yet another transformation stage in grundsteuer which would introduce additional complexity and become a potential source of bugs. Personally, I would leave that for now and maybe revisit it once we start to implement validation.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: Right. Will fix.

tests/samples/grundsteuer_sample_data.py
shrzaf: Remind me to revisit this later. If there's still time once I'm done with grundstueck, gladly.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
```diff
@@ -7,12 +7,256 @@
...
+from erica.request_processing.erica_input.v2.grundsteuer_input_eigentuemer import Vertreter, Anteil, Person, \
+    Empfangsbevollmaechtigter, Eigentuemer
+from erica.request_processing.erica_input.v2.grundsteuer_input_gebaeude import WeitereWohnraeumeDetails, GaragenAnzahl
 from tests.samples.grundsteuer_sample_data import get_sample_vertreter_dict, get_sample_single_person_dict, \
-    get_grundsteuer_sample_data, get_sample_empfangsbevollmaechtigter_dict
+    get_grundsteuer_sample_data, get_sample_empfangsbevollmaechtigter_dict, SampleGebaeude
+
+
+class TestEWohnUnter60:
+    def test_flaeche_under_60_should_set_fields(self):
```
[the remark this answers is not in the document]
shrzaf: Done.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: No, for the same reasons we discussed privately. To summarize: any additional attributes set in the constructor will not be returned by `fields()` (which is used under the hood by `asdict()`) and hence will not end up in the XML. That was the only reason why I considered counting the variables useful before.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: I purposefully do not test for the contents of the object here since that part is tested in `class TestEWohnUnter60`. That way, testing the calculation of the contents of Wohn_Unter60 instance is separated from its assignment in EAngDurchschn, where only the latter is a concern of this test. Does that make sense?

erica/elster_xml/grundsteuer/elster_data_representation.py
```diff
@@ -3,15 +3,146 @@
```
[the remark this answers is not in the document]
shrzaf: It might make sense to split this as well but I wouldn't do it *just* because we split grundsteuer_input. This is a separate "package" with its own unique set of conditions and its structure should not be tightly coupled to that of another package. 

Still, it might be a good idea to split this on its own merit. I will try something and commit it if there's no reason not to.

erica/request_processing/erica_input/v2/grundsteuer_input_gebaeude.py
shrzaf: > Why did you decide on a root validator instead of

Because we are validating a field which cannot be validated independently, but  rather depends on another field having been already parsed. If `ab1949` has not already been parsed by the time `baujahr` is being parsed (and validated), we either get false results or an error.

> Also the error message should be easier understandable, because the field is actually included in it. Not sure if this is also the case for a root_validator.

Right, and great catch. Will fix the messages right away.

erica/request_processing/erica_input/v2/grundsteuer_input_gebaeude.py
shrzaf: Brilliant idea! Those optionals are actually wrong in the first place...

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: Right, I actually missed this. Tests are updated.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: How did you even... took me about a full minute to find it even with your explicit remark :D

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: Because either `wohnflaeche` or `wohnflaechen` must be present in the input dict, otherwise Gebaeude construction* fails on validation.

* no pun intended

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: To make it explicit and visible what's going on here. We construct a valid input object with the garage flag and anzahl set, then explicitly set the flag to false to create the invalid state we'd like to test.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: That we're not testing the instance's state? See my comments above... unless I misunderstood.

tests/request_processing/erica_input/v2/test_grundsteuer_input.py
shrzaf: Yes, thank you, forgot to update this after the rename. Fixed.

tests/request_processing/erica_input/v2/test_grundsteuer_input.py
shrzaf: Right, this is another remnant from before a rename... the test is redundant now.

erica/request_processing/erica_input/v2/possibly_aliased_input.py
shrzaf: Good idea, I will rename it to CamelCaseInput.

We do need `allow_population_by_field_name` though since otherwise snake-cased fields won't be parsed.

erica/elster_xml/grundsteuer/elster_data_representation.py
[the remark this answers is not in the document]
shrzaf: Done.

erica/elster_xml/grundsteuer/elster_data_representation.py
shrzaf: You make a good point. The three EWohn* instances do still just get the information they need to construct their own instances. The limitation here is that we do not want to update the instances in EAngDurchschn as long as the anzahl/flaeche field in each of the three constructed EWohn* instances is 0 (so that the instances with 0 will remain `None` pointers); otherwise they will get serialized into the XML. So we're forced to live a little bit of leakage here in EAngDurchschn. Does that make sense? Or maybe I don't understand your suggestion. Let's talk about it in person if that's the case.

erica/elster_xml/grundsteuer/elster_data_representation.py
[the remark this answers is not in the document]
frederike-ramin: I don't mind them too bad, so we can leave them.

erica/elster_xml/grundsteuer/elster_data_representation.py
frederike-ramin: Lat's talk about it.

erica/elster_xml/grundsteuer/elster_data_representation.py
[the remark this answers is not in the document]
frederike-ramin: I thought *you* were always the person to tell me that we can optimize the input and have the translation layer anyways. 😂 Plus, just basically pasting this code in the webapp can spare us quite some work in erica going forward (introducing more wohnungen for example). So this to me is a very needed translation. (Btw, we already ned a translation layer anyway, because we need to translate "true"/"false" to boolean.

erica/request_processing/erica_input/v2/grundsteuer_input_gebaeude.py
frederike-ramin: As long as you write the field you are checking after the field you are comparing it to in the declaration, there shouldn't be any problem. I just find the specific validators much easier to read, not only because of the error message.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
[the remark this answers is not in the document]
frederike-ramin: Thank you! 🙏

tests/elster_xml/grundsteuer/test_elster_data_representation.py
frederike-ramin: Mh...But I could have by accident added an additional field to the class' definition and this test wouldn't catch that..

tests/elster_xml/grundsteuer/test_elster_data_representation.py
frederike-ramin: It does make sense, but I still am not sure that it's set correctly. I just think it doesn't hurt to explicitly test the content and it'd give me a higher confidence.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
frederike-ramin: Hahaha, nice one! 😂 
Ah, so the factory doesn't make sure it's a valid input? I expected it to.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
frederike-ramin: Mh..although this would be an invalid input to begin with? I feel like we don't need to and even shouldn't test that. I'd like to rely on the input being valid (with regards to the input validators only, of course).

What might be interesting: What if I input "garagen: true, anzahl: 0"? Do we catch that anywhere? Mh...but even if not, we'll get an error from Eric, so it should be fine. But we should think about that in the webapp.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: Done.

erica/elster_xml/grundsteuer/elster_data_representation.py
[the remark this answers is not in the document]
frederike-ramin: Talked about it, we'll revisit in case we have time.

erica/elster_xml/grundsteuer/elster_data_representation.py
frederike-ramin: This would be my suggestion:

```

@dataclass
class EWohnUnter60:
    E7403131: int = 0  # Anzahl
    E7403132: int = 0  # Flaeche

    def __init__(self, flaechen: List[int]):
        for flaeche in flaechen:
            self.E7403131 += 1
            self.E7403132 += flaeche


@dataclass
class EWohn60bis100:
    E7403141: int = 0
    E7403142: int = 0

    def __init__(self, flaechen: List[int]):
        for flaeche in flaechen:
            self.E7403141 += 1
            self.E7403142 += flaeche


@dataclass
class EWohnAb100:
    E7403151: int = 0
    E7403152: int = 0

    def __init__(self, flaechen: List[int]):
        for flaeche in flaechen:
            self.E7403151 += 1
            self.E7403152 += flaeche


@dataclass
class EWeitereWohn:
    E7403121: int = 0
    E7403122: int = 0

    def __init__(self, weitere_wohnraeume_details: WeitereWohnraeumeDetails):
        self.E7403121 = weitere_wohnraeume_details.anzahl
        self.E7403122 = weitere_wohnraeume_details.flaeche


@dataclass
class EAngDurchschn:
    Wohn_Unter60: Optional[EWohnUnter60] = None
    Wohn_60bis100: Optional[EWohn60bis100] = None
    Wohn_ab100: Optional[EWohnAb100] = None
    Weitere_Wohn: Optional[EWeitereWohn] = None

    def __init__(self, input_data: GebaeudeInput):
        flaechen: List[int] = []
        if input_data.wohnflaeche:
            flaechen.append(input_data.wohnflaeche.wohnflaeche)
        elif input_data.wohnflaechen:
            flaechen.append(input_data.wohnflaechen.wohnflaeche1)
            flaechen.append(input_data.wohnflaechen.wohnflaeche2)

        flaechen_unter60 = []
        flaechen_60bis100 = []
        flaechen_ab100 = []
        
        for flaeche in flaechen:
            if flaeche < 60:
                flaechen_unter60.append(flaeche)
            elif flaeche < 100:
                flaechen_60bis100.append(flaeche)
            else:
                flaechen_ab100.append(flaeche)
        
        self.Wohn_Unter60 = EWohnUnter60(flaechen_unter60)
        self.Wohn_60bis100 = EWohn60bis100(flaechen_60bis100)
        self.Wohn_ab100 = EWohnAb100(flaechen_ab100)

        if input_data.weitere_wohnraeume.has_weitere_wohnraeume:
            self.Weitere_Wohn = EWeitereWohn(input_data.weitere_wohnraeume_details)

```

erica/request_processing/erica_input/v2/grundsteuer_input_gebaeude.py
shrzaf: Done.

tests/elster_xml/grundsteuer/test_elster_data_representation.py
shrzaf: Done.

erica/elster_xml/grundsteuer/elster_data_representation.py
shrzaf: Done.

frederike-ramin (approved):
Thank you so much for all the adaptions! 🙏 
I think it's just great!

PR #23, opened 2022-03-17
Add background job tests

JulianRoesner:
# Short Description
This adds tests for the background job. While writing the tests, I noticed that some details were missing and I have added some implementation of missing methods. 

# Changes
- Add tests for background job
- Add additional functionality to the background job worker (e.g. returning the job)
- Remove duplicated coded in lib/pyeric and erica/domain/ElsterXML. We decided to keep that in the `erica_legacy` folder and settle on a permanent placement in the following days.

# Feedback
- What do you think about the way the tests are done?
- I did not want to just add MagicMock and check for the calls but if you think this would be better, we could either write our own queueMock or exchange that with a MagicMock. WDYT?
- Do you think we could get rid of the scheduled_enqueue function for now?
- Did I miss some of the duplicated code or did I miss some imports?


punknoir101 (approved):

SannyNguyenHung (approved):
LGTM! :)

PR #24, opened 2022-03-17
Add tests for base repository

JulianRoesner:
# Short Description
This adds tests for base repository. Furthermore, I added the raising of an error if an entity could not be found in the database.

# Changes
- Added tests for base repository
- Added tests for the InfrastructureModule
- Removed the injection in the init of the Repository as this is not needed.

# Feedback
- Do you see any problems with the implementation?
- Do you see any missing tests?
- I wanted to the test the injections using the infrastructure module. However, I could not get the interface binding to work. Any ideas why that is?


tests/infrastructure/sqlalechemy/repositories/test_base_repository.py
```diff
@@ -0,0 +1,198 @@
...
+        schema_object = MockSchema(**mock_object.dict(), job_id=job_id)
+        transactional_session.add(schema_object)
+        transactional_session.commit()
+
+        found_entity = MockBaseRepository(db_connection=transactional_session).get_by_job_id(job_id)
+
+        assert found_entity == mock_object
+
+    def test_if_entity_not_in_database_then_raise_exception(self, transactional_session):
+        mock_object = MockDomainModel(payload={'endboss': 'Melkor'})
```
punknoir101, line 138: mock_object not used

tests/infrastructure/sqlalechemy/repositories/test_base_repository.py
```diff
@@ -0,0 +1,198 @@
...
+
+        assert len(transactional_session.query(MockSchema).all()) == 0
+
+    def test_if_entity_not_in_database_then_raise_error(self, transactional_session):
+        mock_object = MockDomainModel(payload={'endboss': 'Melkor'})
+        schema_object = MockSchema(**mock_object.dict())
+
+        with pytest.raises(EntityNotFoundError):
+            MockBaseRepository(db_connection=transactional_session).delete(schema_object.id)
+
```
punknoir101, line 198: Wouldn't it be enough to test the erica repository?

erica/infrastructure/sqlalchemy/repositories/erica_request_repository.py
```diff
@@ -0,0 +1,20 @@
...
+from erica.infrastructure.sqlalchemy.erica_request_schema import EricaRequestSchema
+from erica.infrastructure.sqlalchemy.repositories.base_repository import BaseRepository
+
+
+class EricaRequestRepository(
+    BaseRepository[EricaAuftrag, EricaRequestSchema],
+    EricaRequestRepositoryInterface,
+    ABC
+):
+    def __init__(self, db_connection: Session):
```
punknoir101, line 17: You have removed the injection here - why?

erica/infrastructure/sqlalchemy/repositories/erica_request_repository.py
```diff
@@ -0,0 +1,20 @@
+from abc import ABC
+
+import injector as injector
```
punknoir101, line 3: If you remove the injection you can remove this import as well.

erica/infrastructure/sqlalchemy/repositories/erica_request_repository.py
JulianRoesner: When we inject the db_connection in the InfrastructurModule we do not need it here. Furthermore, when we inject as a default value this leads to some problems:
- It's harder to mock it because, when we use the ericaRequestRepository in the tests without an opyoid configuration this will always set the connection to the one implied here. This lead to some errors in the code
- In python the default value can act as a singleton if it is an object. For the session that is fine but for other parts this can be problematic

tests/infrastructure/test_infrastructure_module.py
```diff
@@ -0,0 +1,14 @@
...
+from erica.infrastructure.sqlalchemy.repositories.erica_request_repository import EricaRequestRepository
+
+
+class TestInfrastructureModule:
+
+    def test_if_used_to_inject_erica_req_repository_then_db_connection_correct_injected(self):
+        injector = Injector([InfrastructureModule()])
+        erica_request_repository = injector.inject(EricaRequestRepository)
+
+        assert erica_request_repository.db_connection.bind == DatabaseSessionProvider().get().bind
```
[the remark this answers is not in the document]
JulianRoesner: Yes, this would be tested via integration, too. However, I think there are two reasons why I would test that here:

- It makes it easier to debug. If the integration test fails, it takes time to figure out why something does not work. These tests give us the confidence that using the InfrastructureModule everything is correctly injected and that changes to some parts of the infrastructure does not lead to any problems with this. 
- Tests should always also be a way to document how we intend that our code is used. The usage of the opyoid library is not really straightforward. It took me quite some time (and these tests tbh) to figure out how it works. I would hope that this makes that more easy

erica/infrastructure/sqlalchemy/repositories/base_repository.py
```diff
@@ -30,19 +37,34 @@ def get(self, skip: int = 0, limit: int = 100) -> List[T]:
...
+        return entity
 
     def update(self, entity_id: UUID, model: BaseModel) -> T:
-        current = self.db_connection.query(self.DatabaseEntity).filter(self.DatabaseEntity.id == entity_id)
+        current = self._get_by_id(entity_id)
         current.update(model.dict())
         self.db_connection.commit()
-        updated = self.db_connection.query(self.DatabaseEntity).filter(self.DatabaseEntity.id == entity_id).first()
+
+        updated = self.get_by_id(entity_id)
```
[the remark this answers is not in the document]
JulianRoesner: Thanks for catching that. It is actually the wrong method because we should use the `get_by_job_id` method here.

tests/infrastructure/sqlalechemy/repositories/test_base_repository.py
JulianRoesner: Yes, we could also do that. However, as the implementation is in the base repository, imo it made more sense to test the base repository and then only add tests for special implementation of the erica repository

Pipfile
```diff
@@ -34,6 +34,7 @@ flake8 = "*"
 invoke = "*"
 xmldiff = "*"
 requests = "*"
+pytest-frozen-uuids = "*"
```
SannyNguyenHung, line 37: We don't use that library or?

SannyNguyenHung (commented):
LGTM! Just one comment :+1:

SannyNguyenHung (approved):

PR #27, opened 2022-03-18
Add generic JobService

JulianRoesner:
# Short Description
This adds a generic job service that can be used with the JobServiceFactory to get a service for the different jobs. The jobs are wired up using injection.
- [x] Removing the already implemented Services

# Changes
- Add generic job + interface
- Add tests for the generic job
- Add the service factory that correctly wires the different parts of a service together

# Feedback
- Which parts of the code would you like feedback on?
- Questions that are open are also good (e. g. "Do you see any tests missing for class x?")


erica/application/JobService/job_service_factory.py
```diff
@@ -0,0 +1,41 @@
...
+from erica.application.FreischaltCode.Jobs.jobs import activate_freischalt_code
+from erica.application.JobService.job_service import JobService, JobServiceInterface
+
+injectors = []
+
+
+def _freischalt_code_activation_injector(dto: FreischaltCodeActivateDto):
+    _freischalt_code_activation_injector.handle_type = FreischaltCodeActivateDto
+    
+    module = ApiModule()
```
punknoir101, line 18: Layered architecture shouldn't be affected, what means, the application layer shouldn't know anything about the api layer.

punknoir101:
I think that are several solid principles could be violated.
Anyway, the layered architecture shouldn't be affected, what means, the application layer shouldn't know anything about the api layer.

erica/application/JobService/job_service_factory.py
JulianRoesner: You mean because of the ApiModule? We could also use the ApplicationModule here instead or should we move the service factory to the api layer?

SannyNguyenHung: fixed

tests/application/job_service/test_job_service.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @freeze_time("2001-01-03 08:22:00")
+    @pytest.mark.freeze_uuids
+    def test_if_input_data_provided_then_add_request_to_repository(self):
+        mock_job = PickableMock()
+        service = JobService(job_repository=MockEricaRequestRepository(), background_worker=MagicMock(),
+                             request_controller=MockRequestController, payload_type=MockDto, job_method=mock_job)
+        input_data = MockDto.parse_obj({'name': 'Batman', 'friend': 'Joker'})
+
+        service.apply_queued_to_elster(input_data, job_type=AuftragType.freischalt_code_activate)
```
frederike-ramin, line 83: Hi! Just a quick note - I really misunderstood this method. `apply_queued_to_elster` sounds like executing the job directly. And this is write the opposite of what is being done here - merely queuing the job. Therefore, I'd very much prefer a renaming to `add_to_queue` `add_to_elster_job_queue` `add_as_elster_job_to queue` or something along the lines of this.

JulianRoesner: then `add_to_queue`

erica/application/JobService/job_service_factory.py
```diff
@@ -0,0 +1,69 @@
...
+}
+
+
+
+def get_job_service(request_type: RequestType) -> JobServiceInterface:
+    """
+    This is a factory to get a corretly wired job service. Use that function to get any JobServiceInterface instance.
+
+    :param request_type: The request type. The JobServiceInterface is chosen and wired based on this type.
+    :param dto: Optional data transfer object; the input coming in through the API.
```
VictorDelCampo, line 61: Remove non-existent parameter from the docs

erica/application/EricaAuftrag/EricaAuftrag.py
```diff
@@ -11,7 +11,7 @@ class BasePayloadDto(BaseModel):
 
 
 class EricaAuftragDto(BaseModel):
```
VictorDelCampo, line 13: We wanted to add an additional class attribute for the "result"?

erica/domain/Shared/BaseDomainModel.py
```diff
@@ -9,12 +9,12 @@
 
 class AuditedModel(BaseModel):
     created_at: datetime.datetime
-    updated_at: datetime.datetime
+    updated_at: Optional[datetime.datetime]
```
VictorDelCampo, line 12: Fine, but it doesn't need to be optional. When the entity is created, the updated value is the same as created and later only the updated field is changed.

erica/infrastructure/sqlalchemy/EricaAuftragSchema.py
```diff
@@ -15,7 +15,7 @@ class EricaAuftragSchema(AuditedSchemaMixin, BaseDbSchema):
     id = Column(UUID(as_uuid=True),
```
VictorDelCampo, line 15: We wanted to change the primary key from a UUID to a normal integer right?

VictorDelCampo (approved):
LGTM! Just small comments

erica/application/EricaAuftrag/EricaAuftrag.py
JulianRoesner: This is part of the other pr. But we could add it here, too.

erica/application/JobService/job_service_factory.py
SannyNguyenHung: Done

erica/infrastructure/sqlalchemy/EricaAuftragSchema.py
JulianRoesner: Also Part of the base repository pr because there are too many changes to that already

erica/application/EricaAuftrag/EricaAuftrag.py
VictorDelCampo: Ok

erica/infrastructure/sqlalchemy/EricaAuftragSchema.py
VictorDelCampo: Alrighty

erica/domain/Shared/BaseDomainModel.py
JulianRoesner: Yes. However, I think that it is better to set the updated_at via SQLAlchemy (basically always updating updated_at, when we update it. Therefore, we would not need to set it in the beginning.


But we could

PR #31, opened 2022-03-22
Add service and job for checking tax number

JulianRoesner:
# Short Description
This adds the job for checking the tax number. I changed the RequestController to resemble the others. I ran the integration tests ✅ 

# Changes
- Change CheckTaxNumberRequestController to have a process methods as all the other controllers
- Added missing tests for the service factory.

# Feedback
- Does it makes sense where I put the StateAbbreviation?
- Wdyt about the changes to the requestController? Does that make sense to you?


erica/application/base_dto.py
```diff
@@ -0,0 +1,5 @@
+from pydantic import BaseModel
+
+
+class BaseDto(BaseModel):
+    pass
```
SannyNguyenHung, line 5: Nice! :+1:

erica/erica_legacy/request_processing/requests_controller.py
```diff
@@ -179,13 +181,13 @@ def generate_json(self, pyeric_response: PyericResponse):
         return response
 
 
-class CheckTaxNumberRequestController:
+class CheckTaxNumberRequestController(EricaRequestController):
```
SannyNguyenHung, line 184: I know this is kind of ugly because we don't make any call to erica here. Should we add a new BaseRequestController or should we ignore it for now?

SannyNguyenHung (approved):
LGTM!

erica/application/base_dto.py
JulianRoesner: I think this was already there but only in another file

erica/erica_legacy/request_processing/requests_controller.py
JulianRoesner: I agree. It is not that nice and this was basically the reason why it was not an EricaRequestController before. However, we call Eric and we also use that if we get an request to erica. Therefore, I think It would be fine to leave it like that for now, wdyt?

PR #33, opened 2022-03-22
Queue/timestamps with sqlalchemy

JulianRoesner:
# Short Description
We do not need to set timestamps for the entities ourselves but let the database handle this. This adds exactly that. However, `onupdate` will only be called if no value is provided. Therefore, we strip all the values from the input to the update function that have not changed.

# Changes
- Use SQLAlchemy to set the correct timestamps

# Feedback
- Do you think the tests are at the correct place or should they be placed somewhere else?


VictorDelCampo (approved):
LGMT!

PR #35, opened 2022-03-22
Add failure status to database entity for failed jobs

JulianRoesner:
# Short Description
This adds a failed status to an entity in the database if a job failed. For now, this happens in the job. However, if we want to use retries, we probably need to move that to a special exception handler.
This also adds the `error_code` and `error_message` to all EricaRequest classes and sets them if a job fails.

# Changes
- Remove retries from job creation
- Change status if an EricProcessNotSuccessfulError occurred in a job

# Feedback
- Do you see any part that needs to be adapted with the error_message & error_code part that is currently missing?
- Do you see any tests that I missed?

### DIFF
--- a/erica/application/EricaAuftrag/EricaAuftrag.py
+++ b/erica/application/EricaAuftrag/EricaAuftrag.py
@@ -1,3 +1,4 @@
+from typing import Optional
 from uuid import UUID
 
 from pydantic.main import BaseModel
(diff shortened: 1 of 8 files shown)


SannyNguyenHung (approved):
LGTM!

VictorDelCampo (changes requested):
You say you added error_code and error_message to all EricaRequest classes and even set them if job failed, but I think you forgot to push that :)

VictorDelCampo (approved):
Now it looks awesome! Thanks :)

SannyNguyenHung (approved):

JulianRoesner:
@punknoir101 I will merge that into main. If you have additional comments I will add those in a separate PR, ok?

PR #36, opened 2022-03-23
Add correct request UUID column

JulianRoesner:
# Short Description
For the EricaRequest we had one column named `job_id`. This was used to reference the request & was then also added as the actual job_id of the job that used that request entity to be run. 
This already lead to some miscommunication and misunderstanding.  Therefore, this column is now renamed to `request_id`. This can be used to reference a request from outside of Erica. Furthermore, the job_id is no longer set explicitly as we do not need to reference the job from the request entity.

# Changes
- Rename column
- Rename all methods and attributes that are related to the request_id
- No longer set the job_id of the job explicitly
- Set the request_id as the parameter for the job

# Feedback
- No tests failed so far. However, we did not reference the entity correctly in the job (we set the entity.id but then used the get_by_job_id). Do you think, we should add additional tests for that?

### DIFF
--- a/erica/application/EricaAuftrag/EricaAuftrag.py
+++ b/erica/application/EricaAuftrag/EricaAuftrag.py
@@ -15,7 +15,7 @@ class EricaAuftragDto(BaseModel):
     type: RequestType
     status: Status = Status.new
     payload: object
-    job_id: UUID
+    request_id: UUID
     result: Optional[object]
     error_code: Optional[str]
     error_message: Optional[str]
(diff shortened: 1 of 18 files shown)


VictorDelCampo (changes requested):
LGMT! But we do have some tests failing where we try to add the old parameter entity_id

JulianRoesner:
>  But we do have some tests failing where we try to add the old parameter entity_id

Yes, I somehow missed those locally

VictorDelCampo (approved):
Perrrfect!

PR #40, opened 2022-03-24
Finalize GSTL XML

frederike-ramin:
# Short Description
- This makes some (hopefully) final adjustments to the xml generation.
- The acceptance test in TestSendGrundsteuer now does not return an error anymore

# Changes
- Differentiate between steuernummer and aktenzeichen: Some bundesländer use the former, most the latter
- Set the BUFA-Nr according to the steuernummer/aktenzeichen
- Fix typo: Wohn_Unter60 -> Wohn_unter60
- Send wirtschaftliche einheit zaehler with a comma instead of a dot
- Use a smaller default grundstueck flaeche
- Fix the order of XML tags, otherwise, Elster will return an error
- Return the PDF for the grundsteuer request

# Feedback
- Do you think the code is sufficiently tested?


frederike-ramin:
Instead opened #41

PR #45, opened 2022-03-25
Add bruchteilsgemeinschaft fields

frederike-ramin:
# Short Description
- This mainly adds the bruchteilsgemeinschaft fields - hopefully the last set of fields we'll need to implement 🙏 

# Changes
- Add the fields
- Make gebaeude optional (+consistently named :D)
- For Bruchteilsgemeinschaften we want to set a flag on empfangsvollmacht ("Der/Die in den Zeilen 22 bis 28 eingetragene Empfangsbevollmächtigte ist ein/e Empfangsbevollmächtigte/r im Sinne von § 183 der Abgabenordnung.") - refer to the comments of this ticket if you want more info: https://steuerlotse.atlassian.net/browse/STL-1971

# Feedback
- The commits are quite good and as I included more that just one thing in this PR, maybe you wanna follow them instead of looking at the code as a whole.


tests/erica_legacy/samples/grundsteuer_sample_data.py
```diff
@@ -284,25 +285,58 @@ def parse(self):
...
+class SampleBruchteilsgemeinschaft(Builder):
+    name: str
+    adresse: SampleAdresse
+
+    def __init__(self):
+        super().__init__()
+        self.name = "Bruchteilsgemeinschaft Hogsmeade"
+        self.adresse = SampleAdresse().plz("08642").ort("Hogsmeade")
+
+    def with_name(self, name: str):
```
shrzaf, line 297: So far the (loose) convention has been the following:

With the `with_` prefix when the field should be set with whatever non-None, nonempty value automatically, without the prefix for when the caller should provide the value. So this should be `name(self, name: str)`, whereas `with_adresse(sef)` below is already fine.

tests/erica_legacy/samples/grundsteuer_sample_data.py
```diff
@@ -284,25 +285,58 @@ def parse(self):
         return Vertreter.parse_obj(self.build())
 
 
+class SampleBruchteilsgemeinschaft(Builder):
+    name: str
```
shrzaf, line 289: You could get rid of this (and the extra line in build() below) and just set `self.dict["name"]` directly if you'd like to just to simplify things a little. But it's a purely stylistic choice.

erica/erica_legacy/elster_xml/grundsteuer/elster_gebaeude.py
```diff
@@ -89,27 +89,27 @@ class EAngWohn:
...
 
-        if gebauede.garagen.has_garagen:
-            self.Garagen = EGaragen(gebauede.garagen_anzahl)
+        if gebaeude.garagen.has_garagen:
+            self.Garagen = EGaragen(gebaeude.garagen_anzahl)
         else:
             self.Garagen = None
 
-        self.Ang_Durchschn = EAngDurchschn(gebauede)
+        self.Ang_Durchschn = EAngDurchschn(gebaeude)
```
shrzaf, line 115: 🙏 🥴

shrzaf (approved):

erica/erica_legacy/elster_xml/grundsteuer/elster_eigentuemer.py
```diff
@@ -124,22 +120,28 @@ class EEmpfangsbevollmaechtigter:
...
+        self.E7404625 = empfangsbevollmaechtigter.adresse.hausnummer
+        self.E7404626 = empfangsbevollmaechtigter.adresse.hausnummerzusatz
+        self.E7404640 = empfangsbevollmaechtigter.adresse.plz
+        self.E7404627 = empfangsbevollmaechtigter.adresse.postfach
+        self.E7404622 = empfangsbevollmaechtigter.adresse.ort
 
         # input_data.telefonnummer might not be set -> handle specifically
-        if hasattr(input_data, "telefonnummer") and input_data.telefonnummer:
-            self.E7412201 = input_data.telefonnummer.telefonnummer
+        if hasattr(empfangsbevollmaechtigter, "telefonnummer") and empfangsbevollmaechtigter.telefonnummer:
```
shrzaf, line 139: I believe neither this nor any of the other hasattr checks are actually necessary. Pydantic is already putting explicit keys in its model with value == None for all absent fields in input, so this never evaluates to false. Also all tests passed after I removed all hasattr calls.

Once (if!) we simplify the input model for eigentuemer and gebaeude, even the second check would not be required here: you could just set the elster field to input_data.telefonnummer regardless of its value. The check would only be required for cases where there's a constructor call.

tests/erica_legacy/samples/grundsteuer_sample_data.py
frederike-ramin: Thanks for the explanation. The problem with that and the reason I didn't do it (besides not knowing the convention. :D) is, that if I name it "name" (hehe) it is shadowed by the attribute "name" set in __init__, which is string and cannot be called. So either I remove the default name (which I don't really want) or I give this method another name. I could also image "with_explicit_name" or something.

tests/erica_legacy/samples/grundsteuer_sample_data.py
frederike-ramin: This actually fixes my problem in the other comment, so thanks! :D

PR #50, opened 2022-03-29
NO_TICKET | Remodeling DTO outputs, delegating to specific services for every use case category

VictorDelCampo:
# Short Description
- Starting point --> Output DTOs were under the API layer
- Ending point --> Output DTOs are under the application layer and the response creation is done by specific services. This way the API layer doesn't know about the DTO output creation

# Changes
- Moved Output DTOs to the application layer, together with their corresponding input DTOs
- Created new services for every use case category, so that the output DTO creation is done in the application layer and not longer in the API layer.
- test_api updated so that the correct mock is done for the repository.

# Feedback
- Wdyt?


VictorDelCampo:
> It's nice that you now want to introduce individual services with interface segregation again to step by step getting back to the code/architecture that we already had at the beginning! Thx!

Yes! This way the API layer wouldn't have to create the response DTOS no more!

punknoir101:
> Yes! This way the API layer wouldn't have to create the response DTOS no more!

Exactly.

punknoir101 (approved):

VictorDelCampo:
@JulianRoesner @SannyNguyenHung could you also have a look?

erica/api/v2/endpoints/est.py
```diff
@@ -40,8 +45,8 @@ async def get_send_est_job(request_id: UUID):
     :param request_id: the id of the job.
     """
     try:
-        erica_request = get_erica_request(request_id)
-        return create_response(erica_request)
+        tax_declaration_service: TaxDeclarationServiceInterface = injector.inject(TaxDeclarationServiceInterface)
```
SannyNguyenHung, line 48: Why do we moved the injector into the class? The injector the send_est should not use the injector directly instead we should use a method or a parameter. Wdyt?

erica/api/v2/endpoints/est.py
```diff
@@ -2,21 +2,26 @@
...
+from erica.application.tax_declaration.TaxDeclarationService import TaxDeclarationServiceInterface
 from erica.application.tax_declaration.tax_declaration_dto import TaxDeclarationDto
 from erica.domain.Shared.EricaRequest import RequestType
 from erica.infrastructure.sqlalchemy.repositories.base_repository import EntityNotFoundError
 
 router = APIRouter()
 
+injector = Injector([
+    ApiModule(),
+])
```
JulianRoesner, line 21: I am not the biggest fan of that. Because if we hardcode it, we do not really have a possibility to change it for e.g. the tests and then have to go back to mocking. I am wondering if there is a way to get that from somewhere else (s. also the comment from Sanny)

erica/api/v2/responses/model.py
```diff
@@ -1,99 +1,48 @@
...
-    404: model_error_request_queue,
-    500: model_error_request_queue}
+    404: model_error_get_from_queue,
+    500: model_error_get_from_queue}
 
 response_model_get_send_grundsteuer_from_queue = {
     200: {"model": EstResponseDto,
           "description": "Job status of a sent est was successfully retrieved from the queue."},
-    500: model_error_request_queue}
+    500: model_error_get_from_queue}
```
JulianRoesner, line 24: Isn't the 404 missing?

erica/api/v2/endpoints/est.py
VictorDelCampo: We had the injector before in the utils class and we were calling from the API endpoints that utils class. But now we have a service interface for every use case and therefore it is (for now) being injected in every endpoint.
But we could maybe moved that service definition to the injector factory?

erica/application/tax_declaration/TaxDeclarationService.py
```diff
@@ -0,0 +1,46 @@
...
+from erica.application.Shared.response_state_mapper import map_status
+from erica.application.tax_declaration.tax_declaration_dto import EstResponseDto, ResultEstResponseDto
+
+
+class TaxDeclarationServiceInterface:
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def get_response_send_est(self, request_id: UUID) -> EstResponseDto:
+        pass
```
JulianRoesner, line 16: If all those Interfaces had one common base class, we could put  `erica_request_service.get_request_by_request_id(request_id)` and error handling for the JobTypeDoesNotMatchEndpointError into that base class and use it for all of the different services. Wdyt?

JulianRoesner (commented):
Nothing that would prevent that from being merged. Just some comments where I would see potential improvements

erica/api/v2/responses/model.py
```diff
@@ -1,99 +1,48 @@
...
+    422: {"model": ResponseErrorDto, "description": "Job could not be submitted to the queue."},
+    500: {"model": ResponseErrorDto, "description": "Job could not be submitted to the queue."}}
 
 response_model_get_send_est_from_queue = {
     200: {"model": EstResponseDto,
           "description": "Job status of a sent est was successfully retrieved from the queue."},
-    404: model_error_request_queue,
-    500: model_error_request_queue}
+    404: model_error_get_from_queue,
+    500: model_error_get_from_queue}
```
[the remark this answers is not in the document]
VictorDelCampo: So much cleaner!

erica/application/FreischaltCode/FreischaltCodeService.py
```diff
@@ -0,0 +1,71 @@
...
+                idnr=erica_request.payload.get("idnr"))
+            return FreischaltcodeRequestAndActivationResponseDto(
+                processStatus=map_status(erica_request.status), result=result)
+        elif process_status == JobState.FAILURE:
+            return FreischaltcodeRequestAndActivationResponseDto(
+                processStatus=map_status(erica_request.status), errorCode=erica_request.error_code,
+                errorMessage=erica_request.error_message)
+        else:
+            return FreischaltcodeRequestAndActivationResponseDto(
+                processStatus=map_status(erica_request.status))
```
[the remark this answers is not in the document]
VictorDelCampo: Looks good :)

erica/application/tax_declaration/TaxDeclarationService.py
VictorDelCampo: I have just added a base_service, have a look!

erica/application/tax_number_validation/TaxNumberValidityService.py
```diff
@@ -0,0 +1,45 @@
...
+                is_valid=erica_request.result["is_valid"])
+            return TaxResponseDto(
+                processStatus=map_status(erica_request.status), result=result)
+        elif process_status == JobState.FAILURE:
+            return TaxResponseDto(
+                processStatus=map_status(erica_request.status), errorCode=erica_request.error_code,
+                errorMessage=erica_request.error_message)
+        else:
+            return TaxResponseDto(
+                processStatus=map_status(erica_request.status))
```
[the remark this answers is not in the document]
VictorDelCampo: Tests added

tests/application/tax_declaration/test_tax_declaration_service.py
```diff
@@ -0,0 +1,61 @@
...
+from erica.application.Shared.response_dto import JobState
+from erica.application.tax_declaration.TaxDeclarationService import TaxDeclarationService
+from erica.domain.Shared.EricaRequest import RequestType
+from erica.domain.Shared.Status import Status
+from erica.domain.erica_request.erica_request import EricaRequest
+
+
+class TestTaxDeclarationService:
+
+    @pytest.mark.parametrize("status", [Status.new, Status.scheduled, Status.processing])
```
JulianRoesner, line 15: Love that use of the parametrize 👍 

And thanks for those tests 😍

erica/application/Shared/base_service.py
```diff
@@ -0,0 +1,13 @@
+from uuid import UUID
+from erica.application.erica_request.erica_request_service import EricaRequestService
+
+
+class BaseService:
+    erica_request_service: EricaRequestService
+
```
JulianRoesner, line 7: Could we add a type here, that is None for the base class and would be RequestType.freischaltcode_activate etc. for the other classes. But we can also add that in a separate PR

JulianRoesner (commented):
Nice 👍

erica/application/Shared/base_service.py
VictorDelCampo: What do we need type for in the subclasses?

JulianRoesner: This was just an idea how we couldset the correct request types for all the different request. However, we could also put this as a parameter into the function. Nevertheless, this probably does not make sense to do here but directly when we implement raising an error in case the type was not found.

VictorDelCampo: oh true you we thinking a bit further regarding your PR with the error handling. Let me take over in your PR after we merge this one

VictorDelCampo:
@SannyNguyenHung maybe you could have another quick look? I have removed the injectors from the API lever and introduced a service injector within the application layer. I also refactored the test_api so that they do not depend on a running database no more

erica/application/Shared/service_injector.py
```diff
@@ -0,0 +1,24 @@
...
+    injector = Injector([
+        ApplicationModule(),
+    ])
+
+    switcher = {
+        RequestType.freischalt_code_request: injector.inject(FreischaltCodeServiceInterface),
+        RequestType.freischalt_code_activate: injector.inject(FreischaltCodeServiceInterface),
+        RequestType.freischalt_code_revocate: injector.inject(FreischaltCodeServiceInterface),
+        RequestType.check_tax_number: injector.inject(TaxNumberValidityServiceInterface),
+        RequestType.send_est: injector.inject(TaxDeclarationServiceInterface),
```
SannyNguyenHung, line 22: can you add grundsteuer here?

SannyNguyenHung (approved):
LGTM! But don't forget the grundsteuer, maybe you should merge main into your branch

erica/application/Shared/service_injector.py
VictorDelCampo: Done!

VictorDelCampo:
@SannyNguyenHung I have just integrated the main state into this feature branch and upgraded the grundsteuer API. Tests are also updated. Could you have another quick look?

SannyNguyenHung (approved):

PR #51, opened 2022-03-29
Add queue and worker to docker image

VictorDelCampo:
# Short Description
- Added capability of running both APIs at the same time.
- Adapted entrypoint so that we can conditionally run the API or the worker, or both at the same time

# Changes
- Moved __main__  to the new structure
- Moved __init__ to new structure and add condition depending on whether the dongle functionality (for API v1) should be added. Also added the router for API v2, so both APIs will run in parallel
- Removed old api.py of the v2
- Dockerfile adapted 
- Renamed queue.py to get_queue because of overlapping name of rq module
- Update patch in tests
- Entrypoint.sh extended so that it can be conditionally run

# Feedback
- Any?


erica/config.py
```diff
@@ -28,8 +28,10 @@ class Settings(BaseSettings):
     hersteller_id: str = Field("74931", env='ELSTER_HERSTELLER_ID')
     queue_host: str = Field("localhost", env='QUEUE_HOST')
     queue_port: int = Field(6379, env='QUEUE_PORT')
-    default_queues: List[str] = ['dongle', 'cert', 'common']
+    default_queues: str = 'dongle, cert, common'
```
punknoir101, line 31: List import  not longer needed

punknoir101 (approved):

erica/config.py
VictorDelCampo: List was replaced by a simple string with the values separated with comma, is this way ok?

punknoir101: I see no problems, but the former implementation with list comes from @JulianRoesner  I think.

JulianRoesner: Yes, the change is fine. The former was not to be honest.

I am going to remove the import.

erica/config.py
```diff
@@ -27,8 +26,10 @@ class Settings(BaseSettings):
     elster_datenlieferant: str = Field("PLACEHOLDER_DATENLIEFERANT", env='ELSTER_DATENLIEFERANT')
     hersteller_id: str = Field("74931", env='ELSTER_HERSTELLER_ID')
     queue_url: str = Field("redis://default@localhost:6379/0", env='QUEUE_URL')
-    default_queues: List[str] = ['dongle', 'cert', 'common']
+    default_queues: str = 'dongle, cert, common'
     database_url: str = Field("postgresql://postgres:postgres@localhost/db", env="ERICA_DATABASE_URL")
+    only_api: bool = Field(False, env="RUN_ONLY_API")
```
VictorDelCampo, line 31: Do we need this here? We only use RUN_ONLY_API in the entrypoint.sh right?

PR #57, opened 2022-03-31
Add job integration tests

JulianRoesner:
# Short Description
To have a better confidence that the jobs work in the way they should work, I added integration tests that test the integration of the database with the jobs themselves. I did not add the queue for that because this would only test the functionality of RQ and not our code.
Note: These tests rely on a connection to the Postgres database. I am working on getting rid of that for the tests. However, I did not want to add this in this PR already because this would complicate it.

# Changes
- Add integration tests for freischaltcode, tax_declaration and check_tax_number

# Feedback
- Do you see any problems with the approach?
- I was not sure about the placement so I put those tests next to the unit tests of each job. Wdyt?
- Does the naming make sense to you?


erica/erica_legacy/elster_xml/elster_xml_generator.py
```diff
@@ -267,7 +267,7 @@ def _add_vast_request_xml_nutzdaten(xml_top, user_data):
     spez_recht_antrag_xml = SubElement(nutzdaten_xml, 'SpezRechtAntrag')
     spez_recht_antrag_xml.set('version', '3')
     SubElement(spez_recht_antrag_xml, 'DateninhaberIdNr').text = user_data['idnr']
-    SubElement(spez_recht_antrag_xml, 'DateninhaberGeburtstag').text = user_data['dob']
+    SubElement(spez_recht_antrag_xml, 'DateninhaberGeburtstag').text = user_data['date_of_birth']
```
SannyNguyenHung, line 270: we need to update the webapp request to erica also if you change the data name or?

erica/erica_legacy/request_processing/erica_input/v1/erica_input.py
```diff
@@ -220,12 +220,12 @@ class EstData(BaseModel):
...
     idnr: str
-    dob: date
+    date_of_birth: date = Field(alias='dob')
 
 
 class UnlockCodeActivationData(BaseModel):
     idnr: str
-    unlock_code: str
+    freischalt_code: str = Field(alias='unlock_code')
     elster_request_id: str
```
JulianRoesner, line 230: That's why we do not have to adapt the webapp request. 

We set the alias and this will therefore make the translation for us. You can see that I did not change test_integration.py, which tests v1 of the API.

erica/erica_legacy/elster_xml/elster_xml_generator.py
JulianRoesner: I was afraid of that at first, too. However, we not need to do that. That's why I added the alias in the erica_input.py. I commented the relevant lines below. Maybe have a look there?

PR #61, opened 2022-04-01
Add event logging

JulianRoesner:
# Short Description
This adds the logging of the events described [here](https://digitalservice4germany.atlassian.net/wiki/spaces/STL/pages/237076505/2022-01+Erica+Queue#Monitoring). I added a middleware to make it easy to logg all incoming requests as otherwise only 404s are logged on the production cluster.

# Changes
- Add middleware
- Add logging to service

# Feedback
- Do you see any problem with that?


punknoir101 (approved):
I know it is in the tech spec, but I do not understand, why we need this request logging?

JulianRoesner:
> I know it is in the tech spec, but I do not understand, why we need this request logging?

The idea was to logg all the important steps of the request (to basically see what happens). However, we probably should think about adding an individual id for logging to add more information. Do you see a way to improve the logging?

SannyNguyenHung (approved):
LGTM!

PR #66, opened 2022-04-04
Add error handling 

JulianRoesner:
# Short Description
This adds error handling for not found entities and incorrect request types. I had a look at implementing a special error handler for the Internal server error exceptions. We can still do that. However, we will intercept all exceptions of type Exception. Therefore, we do not have a way to get more detail about the exception.
I added an implementation for that. However I am not so sure if we should keep it.

- [x] We should add some integration tests for the wrong request type when we really raise that error in the code
- [x] I did not want to add raising the wrong type error before #50 was merged

I decided against sending an error for the wrong request type. Instead we can return a redirect and help clients to find the correct endpoint. wdyt?

# Changes
- Add new error type
- Add request type grundsteuer: I added this because I was afraid we would forget the redirect url if we added this later on
- Add exception handlers
 
# Feedback
- What do you think about the exception handlers


erica/api/exception_handling.py
```diff
@@ -0,0 +1,45 @@
...
+             "errorMessage": "The requested entity with id {request_id} was not found."},
+            status_code=404,
+        )
+
+    async def jop_type_mismatch_error(request: Request, exc: RequestTypeDoesNotMatchEndpointError):
+        request_id = request.path_params.get('request_id')
+        redirection_url = job_type_to_endpoint[exc.requested_type]
+        logging.getLogger().info(f"The requested entity {request_id} was requested with the incorrect type {exc.requested_type}. Redirect to {redirection_url}")
+
+        return RedirectResponse(url=redirection_url)
```
VictorDelCampo, line 38: We want to keep the default HTTP code 307 for the redirect right?

erica/__init__.py
```diff
@@ -1,23 +1,26 @@
...
 from prometheus_fastapi_instrumentator import Instrumentator
-from starlette.requests import Request
 
+from erica.api.exception_handling import generate_exception_handlers
 from erica.api.v2.api_v2 import api_router_02
-from erica.erica_legacy.api.api import api_router
 from erica.config import get_settings
+from erica.erica_legacy.api.api import api_router
 from erica.erica_legacy.pyeric.eric import verify_using_stick
+from erica.infrastructure.sqlalchemy.repositories.base_repository import EntityNotFoundError
```
VictorDelCampo, line 13: Why do we need this import?

erica/api/exception_handling.py
```diff
@@ -0,0 +1,45 @@
...
+
+def generate_exception_handlers():
+
+    async def entity_not_found_error(request: Request, exc: EntityNotFoundError):
+        request_id = request.path_params.get('request_id')
+        logging.getLogger().info(f"The requested entity {request_id} is not present in the database.")
+
+        return JSONResponse(
+            {"errorCode": "entity_not_found",
+             "errorMessage": "The requested entity with id {request_id} was not found."},
```
VictorDelCampo, line 29: You forgot the the "f" to build the formatted string literal right?

erica/api/v2/endpoints/est.py
```diff
@@ -22,15 +20,11 @@ async def send_est(est_data_client_identifier: TaxDeclarationDto):
...
-        result = get_job_service(RequestType.send_est).add_to_queue(
-            est_data_client_identifier.payload, est_data_client_identifier.clientIdentifier,
-            RequestType.send_est)
-        return RedirectResponse(url='ests/' + str(result.request_id), status_code=201)
-    # TODO specific exception and correct mapping to JSON error response?
-    except Exception:
-        logging.getLogger().info("Could not send est", exc_info=True)
-        return JSONResponse(status_code=422, content=generate_error_response())
+
+    result = get_job_service(RequestType.send_est).add_to_queue(
```
VictorDelCampo, line 24: We still have in every API use case the responses model:
- For posting a job we have both 422 and 500 with the errorCode/errorMessage JSON
- For getting a job we have both 404 and 500 with the errorCode/errorMessage JSON

The exception handlers you wrote are only meant for the "getting a job" methods so:
- When posting a job unexpected errors would always have the default starlette format right? So no errorCode/errorMessage
- When getting a job the 404 error would have the expected errorCode/errorMessage format but unexpected errors would not fulfill the expected body of the response model.

This means we should:
- Remove the responses model when posting a job (422 and 500) right?
- Remove the 500 response model when getting a job right?

VictorDelCampo (changes requested):
Looks nice! Regarding the redirect for the RequestTypeDoesNotMatchEndpointError I think it is a good idea but I don't think that we would come across this error often right? Maybe when we first implement the new API to the webapp

erica/api/exception_handling.py
JulianRoesner: I would say so. I do not see a problem in that

erica/api/exception_handling.py
JulianRoesner: Good catch

erica/api/v2/endpoints/est.py
JulianRoesner: > The exception handlers you wrote are only meant for the "getting a job" methods so:
No, they are for the whole application. So, whenever the application does not handle an EntityNotFoundError the exception handler is called. The same is true for any Exception (the internal server error comes in)

422 would still occur through the fastApi

JulianRoesner: Therefore, I am not sure if we need to adapt something?

VictorDelCampo: Yes I meant you set that exception handlers for the get-jobs because our EntityNotFoundError can, as the code is right now, only happen in those use cases.
We were using before, as specified in the tech spec, the 422 for the case we had an expected error when posting the job. But now only the EntityNotFoundError (404) and the other one for the wrong request type (default 307) will be handled by your exception handler.
The responses models of the API methods are just definitions about how the responses look like when certain HTTP codes are returned. But for those models to be fulfilled, we have to manually generate the JSON response (meaning adding errorCode and errorMessage)

422 can still ocurrs yes but will have the default FastAPI json response

JulianRoesner: Okay, then I understood you correctly: you would say that we should update those definitions to have the correct api description, right?

- get rid of 404 -> use 307 instead
- only have that for the gets
- get rid of the custom 422 -> we could also overwrite the exception handler for the validation error if we wanted to. But I would say that it is probably better to stay with the fastapi implementation for that one.

Thanks for pointing that out and having an eye for that detail👍🏼

VictorDelCampo: I think we have some bad misunderstanding here. Our current definitions of the response models are:
### POST Job
- For **_HTTP 422_** --> ErrorRequestQueue (meaning errorCode and errorMessage for the JSON response)
- For **_HTTP 500_** --> ErrorRequestQueue
- For the successful response with 201 we have the RedirectResponse object (no problem here)

### GET Job
- For HTTP 200 --> we have the specific respones DTO for every use case, so no problem here.
- For HTTP 404 --> ErrorRequestQueue (this one would be ok with your handler, since we are creating the JSON with the errorCode and errorMessage)
- For **_HTTP 500_** --> ErrorRequestQueue

As FastAPI states in their documentation [here](https://fastapi.tiangolo.com/advanced/additional-responses/): _But for those additional responses you have to make sure you return a Response like JSONResponse directly, with your status code and content._

This means, we are defining an specific response model (errorCode/errorMessage) that we are not fulfilling, since FastAPI will be automatically handling those HTTP errors (the ones marked bold and italic above)

I haven't actually tested what happens if and error is thrown of which we defined a response model but for which we didn't manually create our defined response with errorCode/errorMessage.

So my proposal is simple:
- Remove those response models, for the HTTP codes we are not handling ourselves (422 and 500 for POST and 500 for GET)
or
- Adding another exception handler for at least the 500, but I think you wanted to avoid robbing fast api from catching internal server errors right?

VictorDelCampo (approved):
Can't remove myself, therefore I am approving

VictorDelCampo:
@SannyNguyenHung could you please have a look into this PR? Is the one Julian started and I then took over. I have approved it because I can't remove myself and then added you. Sum up:

- We added global exception handlers for EntityNotFoundError, RequestTypeDoesNotMatchEndpointError and all unexpected Exceptions.
- I updated the response models
- Added integration tests

We have 1 small problem I wasn't able to fix. For the RequestTypeDoesNotMatchEndpointError we return a RedirectResponse and wanted to use the default HTTP 307 but it doesn't work. Somehow FastAPI is not allowing it and furthermore doesn't allow any known 30x HTTP codes (301-308). That is why I set a 309, since all "redirects" begin with the 3xx. It is just a code and since the API is not public, it shouldn't be such a big thing.

Wdyt?

JulianRoesner:
I had a quick look and saw that there is an error in how we construct the locations: we set them with a hardcoded string that does not take the actual endpoint (v2 etc.) into account.
Furthermore, one of the urls in the mapping was incorrectly set by me 🙈 I was able to quickly fix that. We could also think about doing the return of the location param in a similar way. If it does not make sense to you, feel free to change it back.

> We have 1 small problem I wasn't able to fix. For the RequestTypeDoesNotMatchEndpointError we return a RedirectResponse and wanted to use the default HTTP 307 but it doesn't work. Somehow FastAPI is not allowing it and furthermore doesn't allow any known 30x HTTP codes (301-308). That is why I set a 309, since all "redirects" begin with the 3xx. It is just a code and since the API is not public, it shouldn't be such a big thing.

I think the changes should fix that problem? As far as I can see FastApi allows 307 but we will then request the actual url and get the correct result for all the endpoints (being redirected to the correct one).

erica/api/exception_handling.py
```diff
@@ -0,0 +1,69 @@
+import logging
```
SannyNg, line 1: I really like the way how we handle errors now! Good job

SannyNguyenHung, line 1: I really like the way how we handle exceptions now! Good job

SannyNguyenHung (approved):
LGTM! nice job

PR #72, opened 2022-04-08
NO_TICKET | Removed global custom handler RequestValidationError

VictorDelCampo:
# Short Description
- The contract tests for v1 are failing because with the error handling we introduced a custom handler for RequestValidationError, which was needed because otherwise those errors would be catch with the interval_server_error handler (where we are catching all errors)
- For v2 was it ok because it is the new API
- For v1 it was no ok because that API is still relying on the default handler from FastAPI

# Changes
- Removed custom handler for RequestValidationError and added default FastAPI handler.

# Feedback
- Do we need the errorCode/errorMessage response format for v2 in case of a RequestValidationError? In that case we need to specify an specific custom handler only for v2.
- Actually we had already the default handler also for v2 but I had to introduce a custom one because otherwise the RequestValidationError would be catch with our global interval_server_error handler.
- I think it should be fine if we just do it as I proposed, meaning using a custom handler that uses the default FastAPI handler.

### DIFF
--- a/erica/api/exception_handling.py
+++ b/erica/api/exception_handling.py
@@ -1,6 +1,7 @@
 import logging
 
 from fastapi import Request
+from fastapi.exception_handlers import request_validation_exception_handler
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse
 from starlette.responses import RedirectResponse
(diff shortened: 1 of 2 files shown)


JulianRoesner (approved):
Thanks. That is very interesting. 

I think it's fine to do it that way. The only other way would be to check the version of the request and only use fast APIs exception handler for 01 requests. Nevertheless, we can leave it like that imo.

SannyNguyenHung (approved):
LGTM!

PR #80, opened 2022-04-21
Add validation problems

JulianRoesner:
# Short Description
While working on grundsteuer I realized that we do not add the correct errors for validation problems instead we set the error code (a number not a string) and then the error message is just the short descriptor of the error. This changes that. We probably should change the error messages in errors.py in the future

# Changes
- Add validation problems to Global Prüffehler answers
- Fix a small problem where we still did use idnr instead of tax_id_number

# Feedback
- Does this make sense to you?

### DIFF
--- a/erica/application/FreischaltCode/FreischaltCodeService.py
+++ b/erica/application/FreischaltCode/FreischaltCodeService.py
@@ -44,7 +44,7 @@ class FreischaltCodeService(FreischaltCodeServiceInterface):
             result = ResultFreischaltcodeRequestAndActivationDto(
                 elster_request_id=erica_request.result["elster_request_id"],
                 transfer_ticket=erica_request.result["transfer_ticket"],
-                idnr=erica_request.payload.get("idnr"))
+                idnr=erica_request.payload.get("tax_id_number"))
             return FreischaltcodeRequestAndActivationResponseDto(
                 processStatus=map_status(erica_request.status), result=result)
         elif process_status == JobState.FAILURE:
@@ -61,7 +61,7 @@ class FreischaltCodeService(FreischaltCodeServiceInterface):
         if process_status == JobState.SUCCESS:
             result = TransferTicketAndIdnrResponseDto(
                 transfer_ticket=erica_request.result["transfer_ticket"],
-                idnr=erica_request.payload.get("idnr"))
+                idnr=erica_request.payload.get("tax_id_number"))
             return FreischaltcodeRevocationResponseDto(
                 processStatus=map_status(erica_request.status), result=result)
         elif process_status == JobState.FAILURE:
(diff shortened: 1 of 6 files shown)


erica/application/FreischaltCode/FreischaltCodeService.py
```diff
@@ -44,7 +44,7 @@ def _get_base_response_freischaltcode(self, request_id: UUID, request_type: Requ
             result = ResultFreischaltcodeRequestAndActivationDto(
                 elster_request_id=erica_request.result["elster_request_id"],
                 transfer_ticket=erica_request.result["transfer_ticket"],
-                idnr=erica_request.payload.get("idnr"))
+                idnr=erica_request.payload.get("tax_id_number"))
```
VictorDelCampo, line 47: Oh yes! Good eye!

erica/application/JobService/job.py
```diff
@@ -46,8 +46,8 @@ async def perform_job(request_id: UUID, repository: base_repository_interface, s
                 f"Job failed: {entity}. Got Error Message: {e.generate_error_response(True).__str__()}",
                 exc_info=True
             )
-            entity.error_code = e.generate_error_response().get('code')
-            entity.error_message = e.generate_error_response().get('message')
+            entity.error_code = e.generate_error_response().get('message')
```
VictorDelCampo, line 49: A bit confusing that we use error_code for the actual "message" from erica, but we create ourselves that dict with code-message-validationproblems right? So it should be fine, but it is quite hard to read

erica/application/JobService/job.py
```diff
@@ -46,8 +46,8 @@ async def perform_job(request_id: UUID, repository: base_repository_interface, s
...
+            entity.error_message = e.generate_error_response().get('validation_problems', e.generate_error_response().get('message'))
```
VictorDelCampo, line 50: Why not just `e.generate_error_response().get('validation_problems')` ?

erica/erica_legacy/pyeric/eric_errors.py
```diff
@@ -278,7 +278,7 @@ def __str__(self):
 
     @staticmethod
     def get_eric_error_code_message(res_code):
-        return _ERIC_ERROR_MESSAGES.get(res_code, "Unknown error message")
+        return _ERIC_ERROR_MESSAGES.get(res_code, "UNKNOWN_ERROR")
```
VictorDelCampo, line 281: Good, following same capital-letters-pattern as we get from eric

VictorDelCampo (approved):
Looks good and good catch! Just some small remarks

erica/application/JobService/job.py
JulianRoesner: You are right. We should change the e.generate_error_response method I think. However, there is still some dependency with v1 that would take some time to make right. Perhaps have a separate parameter details and call the code number would help?

erica/application/JobService/job.py
JulianRoesner: Because not all errors will have validation problems and I did not want to have a None value here for all the others.

erica/erica_legacy/pyeric/eric_errors.py
JulianRoesner: Yes

PR #83, opened 2022-04-22
STL-2050 | Fix cron jobs

VictorDelCampo:
# Short Description
- Finally the cron jobs for the erica should run

# Changes
- Separated cron jobs, before they were in the same file (what is not possible)
- Corrected cron job definitions
- Added permission magement in the dockerfile for the cron jobs pasted to /etc/cron.d (needed as explained [here](https://unix.stackexchange.com/a/478581))
- Method renamed so that they better match the use case.
- Added env file definition in the entrypoint (needed at least for the ERICA_ENV since it was removed from the cron job definition) and as it is used in the webapp

# Feedback
- I tested them locally and they worked! But should try them also on staging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ RUN chmod -x /etc/cron.d/*
 
 COPY . .
 
-# Get tax office list and ERiC librariescd
+# Get tax office list and ERiC libraries
 RUN env ERICA_BUCKET_NAME=$bucket_name AWS_ACCESS_KEY_ID=$access_key_id AWS_SECRET_ACCESS_KEY=$access_key ENDPOINT_URL=$endpoint_url pipenv run python scripts/load_eric_binaries.py download-eric-cert-and-binaries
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -24,17 +24,20 @@ RUN ln -sf /proc/1/fd/1 /app/cronjob_not_finished_output
...
-
-COPY ./entrypoint.sh /entrypoint.sh
+RUN chown root:root /etc/cron.d/*
+RUN chmod go-wx /etc/cron.d/*
+RUN chmod -x /etc/cron.d/*
 
 COPY . .
 
-# Get tax office list and ERiC libraries
+# Get tax office list and ERiC librariescd
```
SannyNguyenHung, line 40: i think the cd on the end was not intended or?

SannyNguyenHung (approved):
LGTM!

Dockerfile
VictorDelCampo: Ops!

PR #85, opened 2022-04-26
Update Development section in README.md 

mpanne:
Add prerequisites and update paths according to new directory structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,12 +93,12 @@ You also need to obtain a test certificate from ELSTER and place it under `erica
 
 ## Developing 👩‍💻 👨‍💻
 
-Start your docker:
+### Start your docker:
 ```bash
 docker-compose up
 ```
 
-Start the erica API:
+### Start the erica API:
 ```bash
 pipenv shell
 export ERICA_ENV=development
@@ -106,8 +106,9 @@ export SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost/db
 alembic upgrade head
 python3 -m erica 
 ```
+Note: Swagger UI will be available under: http://localhost:8000/docs 
 
-Start a worker:
+### Start a worker:
 ```bash
 pipenv shell
 python3 -m erica.infrastructure.rq.worker [dongle|cert|common]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,7 +69,8 @@ library and place the required library files in a `lib` folder.
 
  - Set the environment variable `ERICA_ENV` to `testing`, `development` or similar.
  - Download `ERiC-35.2.8.0-Linux-x86_64.jar` (or a newer version) from the [ELSTER developer portal](https://www.elster.de/elsterweb/infoseite/entwickler).
- - Place the following files into a `lib` folder in _this directory_ such that it matches the given structure:
+ - Unpack the downloaded jar file
+ - Copy the following library files into `erica/erica_legacy/lib` such that it matches the given structure:
```
SannyNguyenHung, line 73: Thanks :raised_hands:

SannyNguyenHung (approved):
Looking good to me!

README.md
```diff
@@ -44,6 +44,16 @@ MIT-Lizenz lizenziert sind.
...
 
 # For Developers 👩‍💻 👨‍💻
+
+## Prerequisites
+
+The following prerequisites must be installed:
+
+- python 3.9
+- pipenv
+- postgresql
```
VictorDelCampo, line 54: Maybe also specifying that it is the postgresql client that is needed (in ubuntu the package is called postgrtesql-client), meaning only the client needs to be installed and not the whole contrib package (postgresql-contrib).
And in addition also maybe the minimum version needed --> 9.1, see [here](https://pytest-pgsql.readthedocs.io/en/latest/installation.html).

README.md
```diff
@@ -78,23 +89,28 @@ _NOTE_: If you use a Mac, get the corresponding `*.dylib` files
...
 docker-compose up
 ```
 
+### Start the erica API:
 ```bash
+pipenv shell
 export ERICA_ENV=development
-python -m erica 
+export SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost/db 
+alembic upgrade head
```
VictorDelCampo, line 106: I think the alembic migration should not be part of the "Start the erica API". Maybe another category before "Start the erica API" maybe called "Run Alembic migration"? Can also be done in 1 line:
`env ERICA_ENV=development SQLALCHEMY_DATABASE_URI=postgresql://postgres:postgres@localhost/db alembic upgrade head`

VictorDelCampo (approved):
Thanks for the README update! Just some small comments

README.md
mpanne: Good point! Done.

README.md
mpanne: Done!

PR #86, opened 2022-04-26
STL-2159 | Added supervisord and creation of new erica_worker image |…

VictorDelCampo:
# Short Description
- Creation of multiple rq worker using supervisord
- Reduction of sleep time of integration tests v2 to 5 seconds

# Changes
- Added creation of new target in Dockerfile because of usage of supervisord for the VM
- supervisord manages now both the 5 workers and the erica api
- entrypoint now simplified since supervisord can conditionally autostart a program (for our case the worker)
- Updated docker-compose.ci for the new image erica_worker and also renamed the previous wrongly named "web" to "api"
- Reorganized the docker file so that packages are only installed in the specific targets whey they are needed (cron or worker images)
- Logs of child processes of supervisord are automatically redirected to the stdout/stderr of supervisor, meaning we have the logs we we had them before (docker logs)

# Feedback
- Something missing? I tested it locally and they are damm fast the 5 workers! Integration tests are almost every time completed with a sleep time of 2 seconds (set them anyway to 5 seconds sleep)
- IMPORTANT: before this PR is merged into main [this infra PR](https://github.com/digitalservice4germany/steuerlotse-infra/pull/48) should be merged first (because of the new image name erica_worker)

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -1,13 +1,30 @@
 # Note: this docker-compose file is used for building staging/production images on GitHub Actions.
 version: '3.8'
 
-# Define common behavior of web
-x-web:
-  &default-web
+# Define common behavior of api
+x-api:
+  &default-api
   image: erica_api:latest
   build:
     context: .
     target: erica
+    args:
+      bucket_name: ${ERICA_BUCKET_NAME}
+      access_key_id: ${ACCESS_KEY_ID}
+      access_key: ${ACCESS_KEY}
+      endpoint_url: ${ENDPOINT_URL}
+      elster_datenlieferant: ${ELSTER_DATENLIEFERANT}
+      elster_hersteller_id: ${ELSTER_HERSTELLER_ID}
+  expose:
+    - 8000
+
+# Define common behavior of worker
+x-worker:
+  &default-worker
+  image: erica_worker:latest
+  build:
+    context: .
+    target: worker
     args:
       bucket_name: ${ERICA_BUCKET_NAME}
       access_key_id: ${ACCESS_KEY_ID}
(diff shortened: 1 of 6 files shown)


docker-compose.ci.yml
```diff
@@ -18,6 +18,23 @@ x-web:
...
+    target: worker
+    args:
+      bucket_name: ${ERICA_BUCKET_NAME}
+      access_key_id: ${ACCESS_KEY_ID}
+      access_key: ${ACCESS_KEY}
+      endpoint_url: ${ENDPOINT_URL}
+      elster_datenlieferant: ${ELSTER_DATENLIEFERANT}
+      elster_hersteller_id: ${ELSTER_HERSTELLER_ID}
+  expose:
+    - 8000
```
SannyNguyenHung, line 36: Is this right? we should not use 8000 or? The worker and the API should use different ports or?

SannyNguyenHung: We don't even have to expose the port for the workers, do we?

supervisord.conf
```diff
@@ -0,0 +1,32 @@
+[supervisord]
+# needed to avoid CRIT level log
+user=root
```
SannyNguyenHung, line 3: :eyes:

SannyNguyenHung (approved):
LGTM, but we should talk before you merged the branch

docker-compose.ci.yml
VictorDelCampo: The image is called erica_worker but it does not only contain the worker but also the API. This is the image to be used in the VM, which has both the API and the worker definition. On the VM this image is the only one where the property DONGLE_CONNECTED is set to True, meaning the API v1 is only running on the VM.
So we do need to expose the port, is not used in the worker but in the API

SannyNguyenHung (approved):

PR #99, opened 2022-05-06
Use transferticket as one word

frederike-ramin:
# Short Description
- Transferticket is actually one word, so why not export it as one word from Erica?

# Changes
- Adapt the outputted data to use `transferticket` instead of `transfer_ticket`
- Add a check for transferticket (+pdf) in the contract tests (which didn't fail with my changes which I had expected)

# Feedback
- Do you think I should adapt the other snake-case variables, too? (in this PR?)
- Do you think it's be better to change all occurrences of transfer_ticket to transferticket, also in the deeper code (aka "legacy" folder)?
- Do you see a problem with pushing this? Rente is still using the v1, correct? So this should not be a problem? @SannyNguyenHung 

### DIFF
--- a/erica/erica_legacy/api/v1/endpoints/grundsteuer.py
+++ b/erica/erica_legacy/api/v1/endpoints/grundsteuer.py
@@ -13,7 +13,7 @@ router = APIRouter()
 def send_grundsteuer(grundsteuer: GrundsteuerPayload, include_elster_responses: bool = False):
     """
     The Grundsteuer data is validated and then send to ELSTER using ERiC. If it is successful, this should return a 201
-    HTTP response with {'transfer_ticket': str, 'pdf': str}. The pdf is base64 encoded binary data of the pdf
+    HTTP response with {'transferticket': str, 'pdf': str}. The pdf is base64 encoded binary data of the pdf
     If there is any error with the validation, this should return a 400 response. If the validation failed with
     {‘code’ : int,‘message’: str,‘description’: str, ‘validation_problems’ : [{‘code’: int, ‘message’: str}]}
     or a 400 repsonse for other client errors and a 500 response for server errors with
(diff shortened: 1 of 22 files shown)


JulianRoesner (approved):
LGTM!

> Do you think I should adapt the other snake-case variables, too? (in this PR?)

I think it would be better to split those into separate parts. Otherwise this will become quite big very fast.

> Do you think it's be better to change all occurrences of transfer_ticket to transferticket, also in the deeper code (aka "legacy" folder)?

Personally I think yes. However, we should make sure that v1 is still the same. If you could do that that would be perfect

frederike-ramin:
Okay, I adapted all occurrences (I hope).
Could you double-check that I did not destroy the v1 API? 
What I payed attention to:
- do not touch test_routes.py
- do not touch test_integration.py

That should be the most important part, right?

SannyNguyenHung:
> Do you see a problem with pushing this? Rente is still using the v1, correct? So this should not be a problem? @SannyNguyenHung

Should be fine!

erica/erica_legacy/api/v1/endpoints/grundsteuer.py
```diff
@@ -24,7 +24,10 @@ def send_grundsteuer(grundsteuer: GrundsteuerPayload, include_elster_responses:
     """
     try:
         request = GrundsteuerRequestController(grundsteuer, include_elster_responses)
-        return request.process()
+        result = request.process()
+        if "transferticket" in result:
+            result["transfer_ticket"] = result.pop("transferticket")
```
JulianRoesner, line 29: 👍

JulianRoesner:
```suggestion
        if "transferticket" in result:
            result["transfer_ticket"] = result.pop("transferticket", None)
```
Just to not run into some error or is that an error that we want to see?

JulianRoesner (approved):
LGTM! I checked all the endpoints and the transfer_ticket is set all the time. Just one comment about how we extract the value.

We probably should check with rente staging that the transfer ticket is extracted correctly, wdyt?

erica/erica_legacy/api/v1/endpoints/grundsteuer.py
frederike-ramin: Isn't the `if` in the line right before making sure that it is present? O.o

JulianRoesner: Yes, I overlooked that

frederike-ramin:
> We probably should check with rente staging that the transfer ticket is extracted correctly, wdyt?

Yes, let's do that once this is merged. 👍  (Just waiting for #103 to fix the pipeline)

PR #103, opened 2022-05-09
Stl 2159 retry strategy for multiple workers

VictorDelCampo:
# Short Description
- Worker with scheduler
- Retry 3 times with 1 second intervals

# Changes
- Added retry when enqueueing
- Readded 5 workers in the supervisord config
- Added "with_scheduler" to the worker, needed when using property interval in the retry object
- Added new test

# Feedback
- We decided to readd the retries because we are having some problems with the dongle connectivity on staging when the contract tests are run. We were getting sporadically ERIC_CRYPT_E_SC_SLOT_EMPTY or ERIC_CRYPT_E_SC_SIG_KEY on different integration tests. 
- I checked the dongle connection (verify_stick was true) and even restarted the pcscd but problem was not solved
- Running the contract tests for v2 with only 1 worker didn't produce any error
- I decided to use config properties, in case we want to set different values for the retry object in the future, without having to create a new image.
- I also had to manipulate a bit a previous test, since now we are using a Retry object.

### DIFF
--- a/erica/application/JobService/job_service.py
+++ b/erica/application/JobService/job_service.py
@@ -58,7 +58,8 @@ class JobService(JobServiceInterface):
             self.job_method,
             created.request_id,
             ttl=get_settings().ttl_queuing_job_in_sec,
-            retry=Retry(max=3, interval=1)
+            retry=Retry(max=get_settings().queue_retry_repetitions,
+                        interval=get_settings().queue_retry_interval_seconds)
         )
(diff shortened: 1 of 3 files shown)


frederike-ramin:
@VictorDelCampo (I'm really sorry, I'm reopening this, but) If I'm not mistaken, this retries every failed job, correct? However, in some cases we have errors that we do not want to be retried: INVALID_TAX_NUMBER or - even worse, because we only have 5 tries in total: in case the FSC is wrong.
So I think we have to differentiate between expected and unexpected errors. 🙈

PR #110, opened 2022-05-12
Move validation problems into result response

frederike-ramin:
# Short Description
- Before, the validation problems were stored in the `error_message` column of the result table. However, as this had a VARCHAR datatype they were stored like this: `"{"Error 1", "Error 2"}`
- Now, they instead get stored in the `result` column

# Changes
- Store validation_problems as dict in result + return as object
- Adapt test for non-validation errors no not include validation_problems

# Feedback
- Do we have any end-to-end tests that I should adapt to be sure this actually works as intended? Because so far, I think e.g. the ResponseDto is never actually tested?

### DIFF
--- a/contract_tests/test_integration_queue.py
+++ b/contract_tests/test_integration_queue.py
@@ -299,6 +299,72 @@ def full_grundsteuer_data():
 
 
 @pytest.fixture()
+def grundsteuer_data_with_validation_errors():
+    grundsteuer_payload = {
+        "grundstueck": {
+            "typ": "baureif",
+            "adresse": {
+                "strasse": "GST Strasse",
+                "hausnummer": "2",
+                "hausnummerzusatz": "GST",
+                "zusatzangaben": "GST Zusatzangaben",
+                "plz": "12345",
+                "ort": "GST Ort",
+                "bundesland": "BB"
+            },
+            "steuernummer": "09841275756757579",
+            "innerhalbEinerGemeinde": "true",
+            "bodenrichtwert": "123,00",
+            "flurstueck": [
+                {
+                    "angaben": {
+                        "grundbuchblattnummer": "1",
+                        "gemarkung": "2"
+                    },
+                    "flur": {
+                        "flur": "1",
+                        "flurstueckZaehler": "23",
+                        "flurstueckNenner": "45",
+                        "wirtschaftlicheEinheitZaehler": "67.1000",
+                        "wirtschaftlicheEinheitNenner": "89"
+                    },
+                    "groesseQm": "1234"
+                }
+            ]
+        },
+        "eigentuemer": {
+            "person": [
+                {
+                    "persoenlicheAngaben": {
+                        "anrede": "frau",
+                        "titel": "1 Titel",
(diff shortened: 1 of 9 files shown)


erica/application/JobService/job.py
```diff
@@ -45,7 +45,9 @@ async def perform_job(request_id: UUID, repository: base_repository_interface, s
                 exc_info=True
             )
             entity.error_code = e.generate_error_response().get('message')
-            entity.error_message = e.generate_error_response().get('validation_problems', e.generate_error_response().get('message'))
+            entity.error_message = e.generate_error_response().get('message')
```
VictorDelCampo, line 48: I know this was already there before, but why do we call `e.generate_error_response()` several times? Couldn't we just do something like `error_response = e.generate_error_response()` and then extract message and validation problems from there?

VictorDelCampo (approved):
LGMT, just 1 small comment! 

What about the FSC actions? Can they get any validation errors? They can't right?

And surely we would need a new integration test where we maybe send a tax declaration with a validation error (like account holder person_b without defining person_b data?) and then test that the state is FAILED and the validations problems are in the result attribute?

JulianRoesner (approved):
LGTM 🚀.

> Do we have any end-to-end tests that I should adapt to be sure this actually works as intended? Because so far, I think e.g. the ResponseDto is never actually tested?

Yes, you are right. Ideally, we would have some kind of integration tests (similar to the ones for the job) for the api. These should test how api and database work together (does the api return the correct result for some kind of data in the database). We could also add a contract test for grundsteuer and est and check the validation errors, wdyt?

erica/application/JobService/job.py
frederike-ramin: Good remark. 👍

frederike-ramin:
> And surely we would need a new integration test where we maybe send a tax declaration with a validation error (like account holder person_b without defining person_b data?) and then test that the state is FAILED and the validations problems are in the result attribute?

Added precisely that test. Works like a charm (after merging the retry PR #103).

contract_tests/test_integration_queue.py
```diff
@@ -728,6 +794,20 @@ def test_if_get_existing_request_then_return_200_and_response_with_correct_param
...
+        uuid = response.headers['Location'].split("/")[2]
+        sleep(6)
+        response = requests.get(ERICA_TESTING_URL + self.endpoint + "/" + uuid)
+        assert response.status_code == 200
+        assert response.json()["processStatus"] == "Failure"
+        assert response.json()["result"] is not None
+        assert response.json()["result"]["validation_errors"] is not None
+        assert response.json()["errorCode"] is not None
+        assert response.json()["errorMessage"] is not None
+
```
JulianRoesner, line 810: Could we maybe have a similar test for est? I can also add that if you want me to

frederike-ramin: I already added it one commit prior:https://github.com/digitalservice4germany/erica/pull/110/commits/7cae9d0cda6f042d82ef1850017c56dd386b1a70

PR #113, opened 2022-05-12
STL-2159 | Added list of no-retry errors

VictorDelCampo:
# Short Description
- Not all eric errors should be retried

# Changes
- Added a list of no-retryable eric errors

# Feedback
- If a job raises the error at the end of the catch, the job will be retry.
- Since we have "unknown" errors that should be retried, I decided to add a list of errors that should NOT be retried.


erica/erica_legacy/pyeric/eric_errors.py
```diff
@@ -263,6 +264,94 @@
                         **_ERIC_GLOBAL_INITIALISATION_ERRORS, **_ERIC_GLOBAL_ERRORS, **_ERIC_TRANSFER_ERRORS,
                         **_ERIC_CRYPT_ERRORS, **_ERIC_IO_ERRORS, **_ERIC_PRINT_ERRORS}
 
+_ERIC_ERRORS_NO_RETRY = ["ALREADY_OPEN_UNLOCK_CODE_REQUEST", "ELSTER_REQUEST_ID_UNKNOWN", "ERIC_GLOBAL_PRUEF_FEHLER",
```
frederike-ramin, line 267: One small thing: As this list is (only) intended for use outside of this file, the name shouldn't start with an underscore - that is convention for protected attributes/constants.

erica/erica_legacy/pyeric/eric_errors.py
```diff
@@ -263,6 +264,94 @@
...
+                         "ERIC_GLOBAL_LANDESNUMMER_UNBEKANNT",
+                         "ERIC_GLOBAL_BUFANR_UNBEKANNT", "ERIC_GLOBAL_LANDESNUMMER_BUFANR",
+                         "ERIC_GLOBAL_LANDESNUMMER_BUFANR", "NO_MATCHING_IDENTIFIER_FOR_UNLOCK_REQUEST",
```
frederike-ramin, line 270: Could you also delete "NO_MATCHING_IDENTIFIER_FOR_UNLOCK_REQUEST" as an error altogether? We don't use it ever and Julian and I couldn't even reconstruct what this was ever used for. :x

erica/erica_legacy/pyeric/eric_errors.py
frederike-ramin, line 267: I think this is a lot and probably will never be updated. I have two suggestions:
1. Reverse this. Only name errors that we do want retried. - that list is potentially shorter?
2. Only name those errors that we know will occur and do not want retried: "ALREADY_OPEN_UNLOCK_CODE_REQUEST", "ELSTER_REQUEST_ID_UNKNOWN", "ERIC_GLOBAL_PRUEF_FEHLER", "ERIC_GLOBAL_BUFANR_UNBEKANNT", "INVALID_BUFA_NUMBER", "INVALID_TAX_NUMBER", "ERIC_GLOBAL_STEUERNUMMER_UNGUELTIG" and most importantly: "ERIC_TRANSFER_ERR_XML_NHEADER"
3. Reference the complete class of error (like `_ERIC_GLOBAL_ERRORS`) and only remove those that we do want retried - which is basically 1. but different^^

I think I'd probably prefer 2 for simplicity and readability, but we probably had that discussion about which errors to retry for a reason. :x

frederike-ramin (changes requested):
I have a few thoughts on this and - could you maybe add a test that the specified errors are not retried?

erica/erica_legacy/pyeric/eric_errors.py
VictorDelCampo: Oh thanks! Changed :)

erica/erica_legacy/pyeric/eric_errors.py
VictorDelCampo: Done!

erica/erica_legacy/pyeric/eric_errors.py
VictorDelCampo: You are right, I thought it would be better to name those errors that do not need a retry because we had in the confluence list "unknown errors" also listed for a retry, but those errors will be automatically retried since in the perform_job we are only catching "EricProcessNotSuccessful"
This is why I reversed the list and added those listed in confluence that should be retried and yes! The list is much shorter

frederike-ramin (commented):
Nice, that looks way better! Just two suggestions and we're good to go, I think. :)

erica/erica_legacy/pyeric/eric_errors.py
VictorDelCampo: Oh sorry! I will delete it! What about the counter 1-7 of the ERIC_CUSTOM_ERROR_CODES list? Should I reorganize so that we still have the 2 or I just remove that error and we have the 2 missing? I suppose those are the codes we get from ELSTER?

erica/erica_legacy/pyeric/eric_errors.py
```diff
@@ -264,93 +264,11 @@
...
-                         "ERIC_GLOBAL_NICHT_GENUEGEND_ARBEITSSPEICHER", "ERIC_GLOBAL_NO_VERSAND_IN_BETA_VERSION",
-                         "ERIC_GLOBAL_UNKNOWN_PARAMETER_ERROR", "ERIC_GLOBAL_UPDATE_NECESSARY",
-                         "ERIC_IO_FALSCHES_VERFAHREN", "ERIC_IO_MASTERDATENSERVICE_NICHT_VERFUEGBAR",
-                         "ERIC_IO_VERSIONSINFORMATIONEN_NICHT_GEFUNDEN", "ERIC_TRANSFER_ERR_OTHER",
-                         "ERIC_TRANSFER_ERR_SEND", "ERIC_TRANSFER_ERR_SEND_INIT"]
+ERIC_ERRORS_WITH_RETRY = ["ERIC_CRYPT_E_SC_SLOT_EMPTY", "ERIC_CRYPT_E_P11_SLOT_EMPTY", "ERIC_TRANSFER_ERR_PROXYCONNECT",
+                          "ERIC_TRANSFER_ERR_CONNECTSERVER", "ERIC_TRANSFER_ERR_NORESPONSE",
+                          "ERIC_TRANSFER_ERR_TIMEOUT",
+                          "ERIC_TRANSFER_COM_ERROR", "ERIC_TRANSFER_ERR_ENDDATENGROESSE",
+                          "ERIC_TRANSFER_ERR_NOTENCRYPTED"]
```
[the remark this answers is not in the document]
VictorDelCampo: I just copied the errors listed in confluence with a useful retry and these two:

- ERIC_TRANSFER_ERR_NOTENCRYPTED
- ERIC_TRANSFER_ERR_ENDDATENGROESSE

are listed as such and these two:

- ERIC_TRANSFER_ERR_SEND
- ERIC_GLOBAL_INTERNER_FEHLER

are not. BUT you have much more knowledge with erica and if you give me another GO! I will add them :)

VictorDelCampo:
@frederike-ramin I updated the test_job file so that we can test now if jobs with non-retryable errors do not re-raise the EricProcessNotSuccessful and if jobs with retryable errors re-raise it :) What do you think?

erica/erica_legacy/pyeric/eric_errors.py
frederike-ramin: No, the way you did it is correct, I think. We heavily rely on those error codes in the webapp and only check for the code as far as I remember. And the ones in this list actually are custom errors that *we* introduced that are special cases of more general Elster Errors.

erica/erica_legacy/pyeric/eric_errors.py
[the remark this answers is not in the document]
frederike-ramin: You are right, I thought I added the sentence "But probably a lot of thought already went into this" - definitely should have. But when you read the description of the named errors in the API-Referenz of Eric, would you agree that the two additional ones seem beneficial to retry (and the other two don't)? Because, if you agree, let's exchange them.

tests/application/job_service/test_job.py
```diff
@@ -15,7 +15,7 @@ class TestJob:
 
     @pytest.mark.asyncio
     async def test_if_service_raises_error_then_job_raises_error(self):
-        mock_apply_to_elster = MagicMock(side_effect=EricProcessNotSuccessful)
+        mock_apply_to_elster = MagicMock(side_effect=EricProcessNotSuccessful(610201215))
```
frederike-ramin: Why did you add an error code here? And isn't calling the exception here changing how the test works? Because I think having the side_effect as a class is probably different than having an instance?

tests/application/job_service/test_job.py
```diff
@@ -189,3 +201,15 @@ async def test_if_job_ran_successful_then_log_runtime_of_job(self):
 
         assert any("Job running time" in logged_msg[1][0] for logged_msg in info_logger.mock_calls)
         assert any(f"{timedelta(seconds=15)}" in logged_msg[1][0] for logged_msg in info_logger.mock_calls)
+
+
+def get_random_retry_error_code():
```
frederike-ramin, line 206: I completely get why you did that, I'm personally just not a big fan of having randomized test, especially with our already shaky pipeline. Adding another uncertainty maybe isn't the best thing? So I'd probably opt for just testing that SLOT_EMPTY is retried and TRANFER_NHEADER isn't. wdyt?

tests/application/job_service/test_job.py
frederike-ramin: I know why you added the code now. :D

VictorDelCampo: Because the retrying of jobs works like this: if within the job execution an exception is raised, then the job will be retried, otherwise it will be marked as finished. That is why I added this part within the exception-catching in the job.py:
` if entity.error_code in ERIC_ERRORS_WITH_RETRY:
raise
`
Therefore we need an specific error code passed in the EricProcessNotSuccessful, otherwise it will be UNKNOWN_ERROR, which is not in the list for a retry

tests/application/job_service/test_job.py
VictorDelCampo: That was my thought too but then I thought that if we modify the list of retryable errors, then we may missed the one we statically specified in the tests. BUT you are right, we can use two errors that for sure:

- 1 should always be retried like ..SLOT_EMPTY
- 1 should never be retried like ALREADY_OPEN_UNLOCK_CODE_REQUEST
?

frederike-ramin: Yes, great. Let's do it like that.

tests/application/job_service/test_job.py
frederike-ramin: Iiinteresting..so you cant pass a class or an instance of the error you want to be raised. TIL. :) Thank you for explaining. 👍

tests/application/job_service/test_job.py
VictorDelCampo: Done

erica/erica_legacy/pyeric/eric_errors.py
VictorDelCampo: Alrighty! Thanks :)

erica/erica_legacy/pyeric/eric_errors.py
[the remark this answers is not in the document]
VictorDelCampo: - ERIC_TRANSFER_ERR_NOTENCRYPTED: [[610101274] Die Antwortdaten waren nicht PKCS#7-verschlüsselt.
- ERIC_TRANSFER_ERR_ENDDATENGROESSE: [610101260] Im XML-String konnte der Text "</DatenGroesse>" nicht
gefunden werden.

The second one would be definitely a non-retryable error but the first one not so sure, maybe something went wrong with the encryption? But sure if something goes wrong is that we couldn't encrypt at all the message and we would get the other SLOT known error, right?

- ERIC_TRANSFER_ERR_SEND: [610101271] Beim Datenaustausch ist ein Fehler aufgetreten.
- ERIC_GLOBAL_INTERNER_FEHLER: [610001032] Interner Fehler aufgetreten. Details stehen ggf. im Logfile
(eric.log).

These two we could add them to the retryable list jep!

So what about:
- We remove ERIC_TRANSFER_ERR_ENDDATENGROESSE from retryable error list.
- We add the other two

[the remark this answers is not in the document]
frederike-ramin: Great! 💯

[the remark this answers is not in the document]
VictorDelCampo: Done :)

tests/application/job_service/test_job.py
```diff
@@ -14,17 +14,27 @@
...
+    async def test_if_service_raises_raisable_error_then_job_raises_error(self):
+        mock_apply_to_elster = MagicMock(side_effect=EricProcessNotSuccessful(610201215))
         mock_service = MagicMock(apply_to_elster=mock_apply_to_elster)
 
         with pytest.raises(EricProcessNotSuccessful):
             await perform_job(request_id=uuid4(), repository=MagicMock(), service=mock_service,
                               payload_type=MagicMock(), logger=MagicMock())
 
+    @pytest.mark.asyncio
+    async def test_if_service_raises_not_raisable_error_then_job_raises_error(self):
```
frederike-ramin, line 26: If you want to, rename this to `test_if_service_raises_not_raisable_error_then_job_does_not_raise_error` if not, no problem, I just saw it.

frederike-ramin (approved):
Thank you so much fr revisit this! 🚀

tests/application/job_service/test_job.py
JulianRoesner, line 26:
```suggestion
    async def test_if_service_raises_not_raisable_error_then_job_does_not_raise_error(self):
```

I can directly fix that

JulianRoesner: I've changed that

PR #119, opened 2022-05-13
Run cron jobs every minute

JulianRoesner:
# Short Description
To ensure that we file early enough the cron job should run more often. Before it could happen that it would take 7 minutes for a job to fail (not being processed (2 minutes waiting time), cron job will run after 5 minutes)

# Changes
- Reduce waiting time

# Feedback
- Do we need to adapt the other cron job, too?
- Does that make sense?


VictorDelCampo (approved):
Awesome! Thanks!

PR #122, opened 2022-05-16
Camel case input and output of Erica

JulianRoesner:
# Short Description
As we are now not only communicating with other python applications, it does make more sense to use camelCase for the JSON data that we receive and send from our application.

# Changes
- All input parameters and output parameters are now camelCase
- Input parameters should also be accepted as snake_case
- There was one place where we still returned idnr instead of tax_id_number. I changed that.

# Feedback
- Do you think we need to add any tests?


erica/application/base_dto.py
```diff
@@ -1,5 +1,5 @@
-from pydantic import BaseModel
+from erica.application.grundsteuer.camel_case_input import CamelCaseInput
```
frederike-ramin, line 1: Can we maybe move the CamelCaseInput here - it now doesn't only concern Grundsteuer anymore. Or maybe even merge BaseDto and CamelCaseInput into the same thing?

erica/application/tax_declaration/tax_declaration_dto.py
```diff
@@ -6,16 +6,24 @@
 from erica.erica_legacy.request_processing.erica_input.v1.erica_input import FormDataEst, MetaDataEst
 
 
+class FormDataEstDto(FormDataEst, BaseDto):
+    pass
+
+
+class MetaDataEstDto(MetaDataEst, BaseDto):
+    pass
+
```
frederike-ramin, line 15: Is this just for naming and camelcasing? I'm a bit confused honestly. :D If it is, maybe add a comment? (I suppose it's because we don't want to change V1?

erica/application/tax_number_validation/check_tax_number_dto.py
```diff
@@ -2,6 +2,7 @@
 
 from erica.application.Shared.response_dto import ResponseBaseDto
 from erica.application.base_dto import BaseDto
+from erica.application.grundsteuer.camel_case_input import CamelCaseInput
```
frederike-ramin, line 5: Not necessary?

tests/api/test_api.py
```diff
@@ -96,11 +96,11 @@ async def test_if_get_fsc_revocation_job_returns_success_status_with_result():
         mock_service = MagicMock(get_request_by_request_id=MagicMock(return_value=erica_request))
         get_service_mock.return_value = FreischaltCodeService(service=mock_service)
         response = await get_fsc_revocation_job(request_id)
-        assert response.processStatus == JobState.SUCCESS
-        assert response.result.idnr == tax_id_number
+        assert response.process_status == JobState.SUCCESS
+        assert response.result.tax_id_number == tax_id_number
```
frederike-ramin, line 100: So the api will not return camelCase? Or is that even one more level up? (And which would be that level? 🤔 )

frederike-ramin (commented):
I think I'm missing something. At which point is the snake_case converted to camelCase? I feel like we can maybe test that part, too? (Or I was overlooking that, that could be)

tests/api/test_api.py
JulianRoesner: No, the Api will return camelCase because the response inherit from BaseDto -> CamelCaseInput

erica/application/base_dto.py
JulianRoesner: Yes, both would make sense to me

erica/application/tax_declaration/tax_declaration_dto.py
JulianRoesner: These need to also inherit from BaseDto. Otherwise, they won't be camelCased. As you pointed out, I do not want to change the FormDataEst and I thought that this would make more sense

tests/api/test_api.py
frederike-ramin: Shouldn't this be `assert response.result.taxIdNumber == tax_id_number` then? (Because `response` is the response of the api?

JulianRoesner: As I understood the CamelCaseInput it transforms camelCase to snake_case attributes when constructing the object and it transforms the snake_case attributes to camelCase when being parsed as JSON.

Here we test the API functions but they still return the normal object (snake_case). In the contract tests we still expect camelCase

erica/application/base_dto.py
JulianRoesner: I thought a little bit more about it and I would prefer to keep them separate as we specifically use BaseDto for Dtos while the camelCaseInput is not that clearly only used for dtos, wdyt?

frederike-ramin: Sure. As long as CamelCase is not hidden in Grundsteuer, I'm fine. :)

tests/api/test_api.py
frederike-ramin: Ah right - on Json parsing. Can we maybe add a test that makes that clear? I honestly wouldn't know what to test - maybe the CamelCaseInput directly?

erica/application/grundsteuer/grundsteuer_input_eigentuemer.py
```diff
@@ -32,18 +32,18 @@ class Adresse(CamelCaseInput):
...
 
-class Anteil(CamelCaseInput):
+class Anteil(CamelCaseModel):
     zaehler: str
     nenner: str
 
 
-class Person(CamelCaseInput):
+class Person(CamelCaseModel):
     persoenlicheAngaben: PersoenlicheAngaben
```
VictorDelCampo, line 47: Also snake_case here right?

tests/api/test_api.py
```diff
@@ -154,11 +154,11 @@ async def test_if_get_grundsteuer_job_returns_success_status_with_result():
         mock_service = MagicMock(get_request_by_request_id=MagicMock(return_value=erica_request))
         get_service_mock.return_value = GrundsteuerService(service=mock_service)
         response = await get_grundsteuer_job(request_id)
-        assert response.processStatus == JobState.SUCCESS
+        assert response.process_status == JobState.SUCCESS
         assert response.result.pdf == pdf
         assert response.result.transferticket == transferticket
```
VictorDelCampo, line 159: Should also be snake_case right --> transfer_ticket (to be changed in the ouput dto too)

VictorDelCampo (approved):
Looks great! Keeping snake_case in our python code as of PEP 8 Standard and having camelCase for both request/response bodies! Great work. Just some small comments

tests/api/test_api.py
VictorDelCampo: Maybe manually doing a JSON serialization of the returned object?

tests/api/test_api.py
JulianRoesner: We changed that in #99 because transfer ticket is just one word.

erica/application/grundsteuer/grundsteuer_input_eigentuemer.py
JulianRoesner: Yes, we could change that @frederike-ramin ?

tests/api/test_api.py
JulianRoesner: Yes, let's test the CamelCaseInput directly and do the JSON serialization there?

tests/api/test_api.py
VictorDelCampo: Ah ok I missed that PR. Although it is only 1 word in German and not in English ;)

erica/application/grundsteuer/grundsteuer_input_eigentuemer.py
frederike-ramin: I think so, yes. Let's not change the corresponding element in the Grundsteuer test data for the integration_tests_queue.py, though. This way, we can make sure it still works with the way Grundsteuer webapp is sending the data.

tests/api/test_api.py
frederike-ramin: True. It just makes things a bit easier^^

tests/application/test_base_dto.py
```diff
@@ -0,0 +1,28 @@
...
+        name_of_pet = 'Pickett'
+        resulting_object = CamelCaseModelChild.parse_obj({'name_of_pet': name_of_pet})
+
+        assert resulting_object.name_of_pet == name_of_pet
+
+    def test_if_parsed_from_dict_with_camel_case_attribute_then_transform_to_object_with_snake_case_attributes(self):
+        name_of_pet = 'Pickett'
+        resulting_object = CamelCaseModelChild.parse_obj({'nameOfPet': name_of_pet})
+
+        assert resulting_object.name_of_pet == name_of_pet
```
frederike-ramin, line 20: Just to be very sure: Can we add a test for something like person_a_name (with only one letter)? Just thinking about that pyhumps problem that we once had.

frederike-ramin (approved):
Love it! ❤️

PR #125, opened 2022-05-18
Deploy Erica to OTC

nfelger:
# Short Description
- This extends the pipeline to deploy to both of our environments.

# Changes
- I factored out reusable deploy workflows for otc and syseleven and made the pipeline and production-deploy workflows call these new workflows
- I added extra secrets for queue/db urls, erica hosts, argocd, certpin
- Build-image was broken, so I tried fixing it. But what's on production is quite old (27 days, 38810cc4fa3f0bc822a45223d178f485aaf83647) and I don't think its docker-compose.ci.yml has an erica_worker image yet… So I can't test this at the moment. @VictorDelCampo Is there any plan to do an erica deploy anytime soon?

# Feedback
- This is the first time I've used [reusable GHA workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows). I like it. Do you like it, too? :)
- Anything else you know about that I may have missed


.github/workflows/build-image.yml
```diff
@@ -47,7 +46,8 @@ jobs:
 
     - name: Push images
       run: |
-        for IMAGE in api cron worker; do
-          docker push ${{ env.DOCKER_REGISTRY }}/erica_${IMAGE} --all-tags
-        done
+          for IMAGE in api cron worker; do
+            docker tag erica_${IMAGE}:${{ github.event.inputs.imageTag }} ${{ env.DOCKER_REGISTRY }}/erica_${IMAGE}:${{ github.event.inputs.imageTag }}
+            docker push ${{ env.DOCKER_REGISTRY }}/erica_${IMAGE} --all-tags
+          done
```
JulianRoesner, line 52: 👍

.github/workflows/cicd-pipeline.yml
```diff
@@ -110,61 +110,64 @@ jobs:
...
+    steps:
       - name: Report Deployment
         uses: satak/webrequest-action@v1.2.3
-        env:
-          IMAGE_TAG: ${{ github.run_number }}
         with:
           url: ${{ secrets.METRICS_DEPLOYMENT_WEBHOOK_URL }}
           method: POST
-          payload: '{"project": "Erica", "version_identifier": "${{ env.IMAGE_TAG }}", "environment": "staging", "link": "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"}'
+          payload: '{"project": "Erica", "version_identifier": "${{ github.run_number }}", "environment": "staging", "link": "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"}'
```
JulianRoesner, line 166: Out of curiosity: why did you move the GitHub.run_number out of the env? Just because this step looks identical in most repos

.github/workflows/cicd-pipeline.yml
```diff
@@ -110,61 +110,64 @@ jobs:
...
           headers: '{"Authorization": "Token ${{ secrets.METRICS_WEBHOOK_TOKEN }}"}'
 
-  contract-tests:
-    needs: [deploy-staging]
+  contract-tests-syseleven:
+    needs: [deploy-staging-syseleven]
```
JulianRoesner, line 170: Why do we not run the contract tests for the queue on OTC?

.github/workflows/production-deploy.yml
```diff
@@ -7,56 +7,39 @@ on:
...
+      REGISTRY_USER: ${{ secrets.REGISTRY_USER }}
+      REGISTRY_PWD: ${{ secrets.REGISTRY_PWD }}
+      ERICA_SSH_KEY: ${{ secrets.CI_ERICA_SSH_KEY }}
+      ERICA_HOST: ${{ secrets.ERICA_HOST_OTC_PRODUCTION }}
+      CERT_PIN: ${{ secrets.CERT_PIN_OTC_PRODUCTION }}
+      QUEUE_URL: ${{ secrets.QUEUE_URL_OTC_PRODUCTION }}
+      ERICA_DATABASE_URL: ${{ secrets.ERICA_DATABASE_URL_OTC_PRODUCTION }}
+      DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
+      ARGOCD_SERVER: ${{ secrets.ARGOCD_SERVER }}
+      ARGOCD_PIPELINE_PASSWORD: ${{ secrets.ARGOCD_PIPELINE_PASSWORD }}
```
JulianRoesner, line 45: I am fine with that as the Erica should work for both systems all the time. Is that the reason why you decided to deploy to production to both systems instead of being able to toggle between those two??

.github/workflows/production-deploy.yml
```diff
@@ -7,56 +7,39 @@ on:
...
+      REGISTRY_USER: ${{ secrets.REGISTRY_USER }}
+      REGISTRY_PWD: ${{ secrets.REGISTRY_PWD }}
+      ERICA_SSH_KEY: ${{ secrets.CI_ERICA_SSH_KEY }}
+      ERICA_HOST: ${{ secrets.ERICA_HOST_PRODUCTION }}
+      CERT_PIN: ${{ secrets.CERT_PIN_PRODUCTION }}
+      QUEUE_URL: ${{ secrets.QUEUE_URL_PRODUCTION }}
+      ERICA_DATABASE_URL: ${{ secrets.ERICA_DATABASE_URL_PRODUCTION }}
+      KUBECONFIG_B64: ${{ secrets.CI_KUBECONFIG_B64 }}
+      METRICS_DEPLOYMENT_WEBHOOK_URL: ${{ secrets.METRICS_DEPLOYMENT_WEBHOOK_URL }}
+      METRICS_WEBHOOK_TOKEN: ${{ secrets.METRICS_WEBHOOK_TOKEN }}
```
JulianRoesner, line 27: You specify these here but I do not see them in reusable-deploy-syseleven.yml? 

I think we should a) add the report deployment step back into this workflow or b) add it to the separated actions. I would prefer a because this would only report once. wdyt?

.github/workflows/reusable-deploy-otc.yml
```diff
@@ -0,0 +1,88 @@
...
+        uses: digitalservice4germany/github-actions/argocd-deploy@e5a88da4a7c71de837e1d813ba6d90faf5ca0c83
+        with:
+          environment: ${{ inputs.environment }}
+          version: ${{ inputs.imageTag }}
+          deploying_repo: erica
+          infra_repo: erica-infra
+          deploy_key: ${{ secrets.DEPLOY_KEY }}
+          app: erica-${{ inputs.environment }}
+          argocd_pipeline_password: ${{ secrets.ARGOCD_PIPELINE_PASSWORD }}
+          argocd_server: ${{ secrets.ARGOCD_SERVER }}
```
JulianRoesner, line 71: Is this updating the image tag in the Erica-infra?

JulianRoesner (changes requested):
Very nice. Some questions and I think the report of the production deployment is broken?

.github/workflows/reusable-deploy-otc.yml
nfelger: It should, but I just noticed that it is not doing it in the way that we need it to… I'll have to replace this with some custom steps

.github/workflows/production-deploy.yml
nfelger: Hm, interesting. The thought never crossed my mind. Eventually, we'll want everything to use the OTC erica and we don't want to have to coordinate deployments, so maybe good to learn the habits required to make that work early by always deploying to both? (I.e. backwards-compatible changes only…)

.github/workflows/production-deploy.yml
nfelger: Ah yes, thanks, that's an oversight. At one point I had it inside the syseleven workflow, but then realised we'd only want to report a deployment once both succeeded.

.github/workflows/cicd-pipeline.yml
nfelger: Oh just because I didn't see the value in defining a variable for a value that's used exactly once and all the other values on that line are inlined instead of variables… We should really make this a shared action though, so we can be clearer about what are parameters and what isn't

.github/workflows/cicd-pipeline.yml
nfelger: We need to be inside the cluster for that, and I don't want to poke a hole into the cluster just for that purpose. I've set them up as a post-sync job here: https://github.com/digitalservice4germany/erica-infra/blob/main/manifests/overlays/staging/contract-tests-job.yaml.

Still looking for ways to fail the build or at least notify us when they fail though (currently you have to spot it in argo). Maybe once we have this? https://github.com/digitalservice4germany/platform/issues/8

JulianRoesner (approved):
LGTM! 🚀

.github/workflows/reusable-deploy-otc.yml
```diff
@@ -54,22 +54,44 @@ jobs:
...
+      - name: Trigger Argo CD sync
+        shell: sh
+        run: |
+          curl -sSL -o argocd https://${ARGOCD_SERVER}/download/argocd-linux-amd64
+          chmod +x argocd
+          ./argocd login ${ARGOCD_SERVER} --username pipeline --password ${{ secrets.ARGOCD_PIPELINE_PASSWORD }} --grpc-web
+          ./argocd app get --refresh erica-${{ inputs.environment }} --grpc-web > /dev/null
+          ./argocd app wait erica-${{ inputs.environment }} --sync --health --grpc-web
+        env:
+          ARGOCD_SERVER: ${{ secrets.ARGOCD_SERVER }}
```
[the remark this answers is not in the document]
nfelger: I copied this from the action @carhartl created. I _think_ this is so we can get a failed build when the sync fails? But I'm not 100% sure what the motivation was.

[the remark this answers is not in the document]
JulianRoesner: Ah, that is cool. Maybe we could also check the contract test jobs through the same mechanism. Why could we not use the action directly? I tried[ it for achill ](https://github.com/digitalservice4germany/achill/blob/0fa384340c71cdf83b229c84aed6f593ceafe82b/.github/workflows/ci-pipeline.yml#L85) and it works fine there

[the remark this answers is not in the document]
nfelger: Two reasons:
- the action only updates a single image (the `deploying_repo`), but erica needs both the `erica_api` and `erica_cron` tags to be updated
- it assumes you use github packages as the image registry, but erica needs (for now - happy to migrate in future) SysEleven SCR

For both, see here: https://github.com/digitalservice4germany/github-actions/blob/main/argocd-deploy/action.yml#L42

I don't think it'd be very hard to update the action to support this, but I didn't have time…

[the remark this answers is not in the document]
JulianRoesner: 🙈 Of course. Thank you very much for clarifying.

PR #150, opened 2022-06-15
Set testmerker for grundsteuer

JulianRoesner:
# Short Description
We need to set the testmerker for grundsteuer request. This does that and enforces that either all tax id numbers are test id numbers or not.

# Changes
- Add tests for the Eigentuemer validation
- Set testmerker if first Eigentümer has test_id_number

# Feedback
- Did I miss something from [our plan](https://steuerlotse.atlassian.net/browse/STL-2324?atlOrigin=eyJpIjoiOTVmMDI5YWY4MDM1NGI3NWI0ZGY4OGZkMTMwODJmZDEiLCJwIjoiaiJ9)?

### DIFF
--- a/erica/application/grundsteuer/grundsteuer_input_eigentuemer.py
+++ b/erica/application/grundsteuer/grundsteuer_input_eigentuemer.py
@@ -83,3 +83,9 @@ class Eigentuemer(CamelCaseModel):
         if not(all([tax_id_number_is_test_id_number(person.steuer_id) for person in v]) or all([not tax_id_number_is_test_id_number(person.steuer_id) for person in v])):
             raise ValueError('all eigentuemer need to use either real or test tax id numbers')
         return v
+
+    @validator("person")
+    def must_not_be_empty(cls, v, values):
+        if len(v) < 1:
+            raise ValueError('at least one eigentuemer needs to be set')
+        return v
--- a/tests/application/grundsteuer/test_grundsteuer_input_eigentuemer.py
+++ b/tests/application/grundsteuer/test_grundsteuer_input_eigentuemer.py
@@ -87,47 +87,50 @@ class TestEigentuemer:
 
     def test_if_all_persons_have_test_tax_id_number_then_raise_no_error(self):
         persons = [Person(persoenliche_angaben=PersoenlicheAngaben(anrede="frau", name="daViella",
-                                                                                vorname="Gruella"),
-                                       adresse=Adresse(plz="79618", ort="Rheinfelden"),
-                                       steuer_id="04452317681", anteil=Anteil(zaehler="1", nenner="2")),
-                                Person(persoenliche_angaben=PersoenlicheAngaben(anrede="herr", name="Man",
-                                                                                vorname="Robin"),
-                                       adresse=Adresse(plz="79618", ort="Rheinfelden"),
-                                       steuer_id="09952417688", anteil=Anteil(zaehler="1", nenner="2")),
-                                Person(persoenliche_angaben=PersoenlicheAngaben(anrede="frau", name="Biest",
-                                                                                vorname="Bella"),
-                                       adresse=Adresse(plz="79618", ort="Rheinfelden"),
-                                       steuer_id="03352417692", anteil=Anteil(zaehler="1", nenner="2"))]
+                                                                   vorname="Gruella"),
+                          adresse=Adresse(plz="79618", ort="Rheinfelden"),
+                          steuer_id="04452317681", anteil=Anteil(zaehler="1", nenner="2")),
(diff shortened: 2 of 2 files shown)


erica/application/EricRequestProcessing/grundsteuer_request_controller.py
```diff
@@ -9,11 +9,14 @@
...
 from erica.erica_legacy.request_processing.requests_controller import TransferticketRequestController
 
+
 class GrundsteuerRequestController(TransferticketRequestController):
     _PYERIC_CONTROLLER = GrundsteuerPyericProcessController
 
     def _is_testmerker_used(self):
-        # TODO revisit this
+        if len(self.input_data.eigentuemer.person) >= 1 and self.input_data.eigentuemer.person[0].steuer_id:
+            print(self.input_data.eigentuemer.person)
```
frederike-ramin, line 18: I don't think we should print this?

erica/application/grundsteuer/grundsteuer_input_eigentuemer.py
```diff
@@ -75,4 +76,10 @@ def must_be_set_only_if_two_persons(cls, v, values):
             raise ValueError('has to be set if two persons')
         if values.get('person') and len(values.get('person')) != 2 and v is not None:
             raise ValueError('must not be set if other than two persons')
+        return v
+
+    @validator("person")
+    def must_have_all_test_ids_or_no_test_ids(cls, v, values):
```
frederike-ramin, line 82: We're not doing this for Rente yet, right? Should we maybe at least point them to the problem?

tests/erica_legacy/request_processing/erica_input/v2/test_grundsteuer_input_eigentuemer.py
```diff
@@ -83,3 +84,56 @@ def test_if_three_persons_and_verheiratet_false_then_raise_error(self):
...
+                                       adresse=Adresse(plz="79618", ort="Rheinfelden"),
+                                       steuer_id="09952417688", anteil=Anteil(zaehler="1", nenner="2")),
+                                Person(persoenliche_angaben=PersoenlicheAngaben(anrede="frau", name="Biest",
+                                                                                vorname="Bella"),
+                                       adresse=Adresse(plz="79618", ort="Rheinfelden"),
+                                       steuer_id="03352417692", anteil=Anteil(zaehler="1", nenner="2"))]
+        try:
+            Eigentuemer.parse_obj({"person": persons})
+        except ValidationError as e:
+            pytest.fail("parse_obj failed with unexpected ValidationError " + str(e))
```
frederike-ramin, line 104: I thought at some point we decided against explicitly failing in tests? Because, at first I was a bit irritated^^

tests/erica_legacy/request_processing/test_grundsteuer_request_controller.py
```diff
@@ -21,7 +21,19 @@ def valid_grundsteuer_request_controller():
...
+
+    def test_if_tax_id_number_no_test_tax_id_then_returns_false(self, valid_grundsteuer_request_controller):
+        valid_grundsteuer_request_controller.input_data.eigentuemer.person[0].steuer_id = "43865766025"
+        result = valid_grundsteuer_request_controller._is_testmerker_used()
+        assert result is False
+
+    def test_if_no_tax_id_number_then_returns_true(self, valid_grundsteuer_request_controller):
+        valid_grundsteuer_request_controller.input_data.eigentuemer.person[0].steuer_id = None
         result = valid_grundsteuer_request_controller._is_testmerker_used()
         assert result is True
```
frederike-ramin, line 38: About this I'm not so sure...is there ever a case this could happen? I think none that makes sense. So, in that case we'd want to get notified I think? Therefore, I'd probably instead throw an error. Returning True still is better than False of course, but maybe an error would make even more sense?

frederike-ramin (approved):
Just a few adjustments and questions - otherwise: very sleek solution. 👌

erica/application/grundsteuer/grundsteuer_input_eigentuemer.py
JulianRoesner: Yes, let's talk to them

tests/erica_legacy/request_processing/test_grundsteuer_request_controller.py
JulianRoesner: I have added a separate validator that ensures that we always have a person. However, the steuer_id is an optional field. Therefore, we do not know if it is set. We talked about it and we'll keep it like that.

frederike-ramin:
@JulianRoesner So Sanny said it would be great to also have the check for Rente. So once we have time again we can implement it for them. :D

PR #154, opened 2022-06-17
Docker compose rework

nfelger:
# Short Description
- Follow-up from our conversation yesterday: removes supervisord and runs only 1 worker.
- This PR has two related PRs:
  * https://github.com/digitalservice4germany/erica-infra/pull/1
  * https://github.com/digitalservice4germany/steuerlotse-infra/pull/57

# Changes for all 3 PRs
- dynamic image tagging in github actions without a hard-coded list of images (`docker-compose.ci.yml` as the single source of truth)
- removed supervisord in favour of docker-compose
- running only 1 worker
- removed `RUN_WITH_WORKER` - worker is now a separate service
- removed setting `DONGLE_CONNECTED=True `in VM deployment script - the default is True, so this setting never did anything
- uses a single image (called erica_main to avoid conflicts with existing images) w/ different CMDs for api and worker; should also reduce build time a little
- removed inconsistent use of `ERICA_DATABASE_URL` and `SQLALCHEMY_DATABASE_URI` (now only `ERICA_DATABASE_URL`)
- added some invoke tasks, tested and simplified setup instructions in the README


.github/workflows/cicd-pipeline.yml
```diff
@@ -103,10 +104,8 @@ jobs:
 
       - name: Push images
         run: |
-          for IMAGE in api cron worker; do
-            docker tag erica_${IMAGE}:${{ github.run_number }} ${{ env.DOCKER_REGISTRY }}/erica_${IMAGE}:${{ github.run_number }}
-            docker tag erica_${IMAGE}:latest ${{ env.DOCKER_REGISTRY }}/erica_${IMAGE}:latest
-            docker push ${{ env.DOCKER_REGISTRY }}/erica_${IMAGE} --all-tags
+          for IMAGE in $(docker image ls | grep erica_ | awk '{ print $1 }' | sort | uniq); do
+            docker push ${IMAGE} --all-tags
```
JulianRoesner, line 108: 👍

README.md
```diff
@@ -65,63 +65,73 @@ pipenv install
...
+
+#### Downloading as a DigitalService employee
+
+You can download binaries and organisation-specific test certificates by running:
+
+```bash
+ERICA_BUCKET_NAME=xxx AWS_ACCESS_KEY_ID=xxx AWS_SECRET_ACCESS_KEY=xxx ENDPOINT_URL=xxx pipenv run download-eric
+```
+
+Contact the erica service owners for information on how to obtain the required credentials and secrets.
```
JulianRoesner, line 78: 👍

tasks.py
```diff
@@ -22,3 +22,19 @@ def lint(c):
...
+def run_api(c):
+    c.run("python -m erica")
+
+@task
+def run_worker(c):
+    c.run("python -m erica.infrastructure.rq.worker")
+
+@task
+def download_eric(c):
+    c.run("python scripts/load_eric_binaries.py download-eric-cert-and-binaries", pty=True)
```
JulianRoesner, line 40: Love it ❤️

docker-compose.yml
```diff
@@ -8,16 +8,13 @@ services:
...
       - POSTGRES_PASSWORD=postgres
+    expose:
+      - 5432
     ports:
       - '5432:5432'
-    volumes:
-      - db:/var/lib/postgresql/data
-volumes:
-  db:
-    driver: local
```
JulianRoesner, line 23: Why are we able to drop those?

docker-compose.ci.yml
```diff
@@ -18,27 +18,10 @@ x-api:
...
-    target: worker
-    args:
-      bucket_name: ${ERICA_BUCKET_NAME}
-      access_key_id: ${ACCESS_KEY_ID}
-      access_key: ${ACCESS_KEY}
-      endpoint_url: ${ENDPOINT_URL}
-      elster_datenlieferant: ${ELSTER_DATENLIEFERANT}
-      elster_hersteller_id: ${ELSTER_HERSTELLER_ID}
-  expose:
-    - 8000
```
JulianRoesner, line 36: Why can we get rid of that?

JulianRoesner (commented):
Just some questions about deleted lines?

docker-compose.ci.yml
nfelger: No separate worker image. Worker uses the main image.

nfelger: Remember this docker-compose file is only used to build images, never to actually run services.

docker-compose.yml
nfelger: Hm, this looks like an accidental commit. But I'd ask the reverse question: why were they there in the first place?

`driver: local` is the default.

And why map the data directory? To keep data between postgres container restarts? Seems odd in a dev environment, but maybe you actually want it to work this way? Then I'd revert.

docker-compose.ci.yml
JulianRoesner: 👍

docker-compose.yml
JulianRoesner: > why were they there in the first place?

That is a good question but I assume that was the reason. But I would agree that especially for the Erica service we do not need to keep data between restarts and it is probably more confusing that we have it.

nfelger:
Merging now, but comments & questions are still welcome! @frederike-ramin @VictorDelCampo

PR #157, opened 2022-06-23
Using huey

JulianRoesner:
# Short Description
Instead of rq we want to use huey. The reason is that we encountered some problems using RQ because we have to use different processes for different workers which does not work with ERiC and a security stick.

# Changes
- Remove RQ
- Add huey component and use it for all the perform_job functions
- Adapt the tests
- Add adr

# Feedback
- I thought that we already had an adr for RQ. Somehow, that was missing. Do you think we should add it now or just go with it and at least have one for huey.
- Does the structure/naming of the huey component make sense to you?
- Do you see any problems that I might have missed.


JulianRoesner:
@VictorDelCampo Maybe you can have a look, too. If you have any questions, let me know.

tasks.py
```diff
@@ -33,7 +33,7 @@ def run_api(c):
 
 @task
 def run_worker(c):
-    c.run("python -m erica.infrastructure.rq.worker")
+    c.run("huey_consumer.py erica.application.JobService.job.huey -k thread -w 10")
```
nfelger, line 36: Probably a very basic question: where does `huey_consumer.py` come from? I didn't see it get added anywhere?

erica/infrastructure/worker/huey_task.py
```diff
@@ -0,0 +1,5 @@
+from erica.config import get_settings
+
+from huey import RedisHuey
+
+huey_task = RedisHuey('erica-huey-queue', url=get_settings().queue_url, immediate=get_settings().use_immediate_worker)
```
nfelger, line 5: Can we call this just `huey`? It's an instance of a RediHuey, so calling it a task confused me. 

Also: did you consider using dependency injection to obtain the Huey instance? Since it's imported at the module level _and_ because we can use the real huey (with the immediate setting), I'm not sure that it would have any benefit at all, I'm just curious if you thought about it.

nfelger: Also the module naming convention seems to be `erica.infrastructure.$libraryName....`. Can we rename this one to `erica.infrastructure.huey`?

nfelger:
> I thought that we already had an adr for RQ. Somehow, that was missing. Do you think we should add it now or just go with it and at least have one for huey.

I'd say what you wrote is detailed enough.

> Does the structure/naming of the huey component make sense to you?

Structure yes, for naming I have commented suggestions. 

> Do you see any problems that I might have missed.

No. 

> Do you see any specific part where we could/need to add documentation for huey?

No.

tasks.py
JulianRoesner: It comes from the [huey package](https://github.com/coleifer/huey/blob/master/huey/bin/huey_consumer.py) that's why it was not added in this PR. I also did not find that very intuitive but it is [the documented way](https://huey.readthedocs.io/en/latest/consumer.html) to start the worker.

erica/infrastructure/worker/huey_task.py
JulianRoesner: > Can we call this just huey? It's an instance of a RediHuey, so calling it a task confused me.

Yes. I changed it to huey_task because I wanted to prevent overloading huey but I

> Also: did you consider using dependency injection to obtain the Huey instance? Since it's imported at the module level and because we can use the real huey (with the immediate setting), I'm not sure that it would have any benefit at all, I'm just curious if you thought about it.

Yes, I thought about it and you are describing my thought process😄. I could not see a good way to use dependency injection without making the code more complicated and because I did not see a clear benefit for either testing or different environments, I decided that the simpler version might be better

> Also the module naming convention seems to be erica.infrastructure.$libraryName..... Can we rename this one to erica.infrastructure.huey

We can.

erica/application/JobService/job_service.py
```diff
@@ -54,18 +49,12 @@ def add_to_queue(self, payload_dto: BaseDto, client_identifier: str, job_type: R
...
-        job = self.background_worker.enqueue(
-            self.job_method,
-            created.request_id,
-            ttl=get_settings().ttl_queuing_job_in_sec,
-            retry=Retry(max=get_settings().queue_retry_repetitions,
-                        interval=get_settings().queue_retry_interval_seconds)
-        )
-        logging.getLogger().info(f"Job created with id {job.id} for EricaRequest with id {created.request_id}")
+        self.job_method(created.request_id)
+        logging.getLogger().info(f"Job created with id for EricaRequest with id {created.request_id}")
```
frederike-ramin, line 53: So I see this correctly that there'll be no retries?

frederike-ramin: In that case we should probably also delete the whole ERIC_ERRORS_WITH_RETRY part?

tasks.py
```diff
@@ -33,7 +33,7 @@ def run_api(c):
 
 @task
 def run_worker(c):
-    c.run("python -m erica.infrastructure.rq.worker")
+    c.run("huey_consumer.py erica.infrastructure.huey.huey.huey -k thread -w 10")
```
frederike-ramin, line 36: We don't want the number of workers to be configurable? (Or maybe `-w` means a completely different thing. :D )

tests/application/job_service/test_job_service.py
```diff
@@ -91,44 +91,35 @@ def test_if_input_data_provided_then_add_request_to_repository(self):
...
-    @pytest.mark.asyncio
-    async def test_if_input_data_provided_then_call_init_of_request_controller_with_correct_data(self):
-        mock_bg_worker = MagicMock()
+    def test_if_input_data_provided_then_call_init_of_request_controller_with_correct_data(self):
         controller_instance = MagicMock()
         mock_request_controller = MagicMock(return_value=controller_instance)
-        service = JobService(job_repository=MockEricaRequestRepository(), background_worker=mock_bg_worker,
+        service = JobService(job_repository=MockEricaRequestRepository(),
                              request_controller=mock_request_controller, payload_type=MockDto, job_method=MagicMock())
         input_data = MockDto.parse_obj({'name': 'Batman', 'friend': 'Joker'})
```
frederike-ramin, line 113: Wellll.......friend?!

frederike-ramin (approved):
Just some minor questions and hints. So happy it works so smoothly!

tests/application/job_service/test_job_service.py
JulianRoesner: I guess that depends on who is looking at that relationship? Joker definitely considers Batman as some kind of a friend in some stories. I added it mostly for fun. So, thanks for finding that one 🤣

PR #165, opened 2022-06-28
Make "import erica" work without hacking sys.path

nfelger:
This is a "SHOW" PR @frederike-ramin @JulianRoesner @VictorDelCampo 

Just a small detail of python packaging that can make our lives easier.

If you make a package installable (by adding a minimal `setup.py`), you can add it to your Pipfile, which will automatically add it to the load path.

More about setup.py and all things packaging here: https://python-packaging.readthedocs.io/en/latest/minimal.html

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,6 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+erica = {path = ".", editable = true}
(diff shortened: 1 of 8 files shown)


JulianRoesner:
That's great. Thanks for explaining, showing and fixing 🙏

PR #169, opened 2022-07-04
Log successful jobs

JulianRoesner:
# Short Description
We do not log successful jobs. If we also want to log the request type we need to set that as a parameter.

# Changes
- Set request type as a parameter to perform job.

# Feedback
- Does that make sense to you?
- Do you see any problem with that being deployed (should be backwards compatible)

### DIFF
--- a/erica/application/JobService/job.py
+++ b/erica/application/JobService/job.py
@@ -6,6 +6,7 @@ from uuid import UUID
 from pydantic import ValidationError
 
 from erica.application.JobService.job_service import JobServiceInterface
+from erica.domain.Shared.EricaRequest import RequestType
 from erica.domain.repositories import base_repository_interface
 from erica.domain.Shared.Status import Status
 from erica.domain.erica_request.erica_request import EricaRequest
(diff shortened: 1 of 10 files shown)


erica/application/JobService/job.py
```diff
@@ -49,6 +50,7 @@ def perform_job(request_id: UUID, repository: base_repository_interface, service
             entity.result = response
             entity.status = Status.success
             repository.update(entity.id, entity)
+            logger.info(f"Job ran successfully: type {job_type}")
```
frederike-ramin, line 53: Like the logging, but don't we log a similar thing (minus the request_type) in l.67?

frederike-ramin: Having a differentiation definitely is the smarter move^^ So, yes, let's delete l.67

JulianRoesner: Good catch. Had another look and l. 67 also logs the job type because this is part of the entity toString method. It also logs the entity status. Therefore, we can close this PR 🙈

JulianRoesner:
The success is already logged

PR #173, opened 2022-07-11
Expose FSC already revoked error to clients

frederike-ramin:
# Short Description
- When implementing the re-requesting of FSCs in Grundsteuer, I was thinking about this and noticed that we ar enot exposing the specific "AlreadyRequestedFSC" error to clients. I think there's no harm in that?

# Changes
- Explicitly add AlreadyRevokedError to possible error codes

# Feedback
- Do you see a reason we shouldn't do this?

### DIFF
--- a/erica/erica_legacy/pyeric/eric_errors.py
+++ b/erica/erica_legacy/pyeric/eric_errors.py
@@ -13,6 +13,7 @@ _ERIC_CUSTOM_ERROR_CODES = {
     5: "ELSTER_REQUEST_ID_UNKNOWN",
     6: "INVALID_BUFA_NUMBER",
     7: "INVALID_TAX_NUMBER",
+    8: "ALREADY_REVOKED",
 }
 
 _ERIC_GLOBAL_VALIDATION_ERRORS = {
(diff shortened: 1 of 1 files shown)


erica/erica_legacy/pyeric/eric_errors.py
```diff
@@ -13,6 +13,7 @@
     5: "ELSTER_REQUEST_ID_UNKNOWN",
     6: "INVALID_BUFA_NUMBER",
     7: "INVALID_TAX_NUMBER",
+    8: "ALREADY_REVOKED",
```
VictorDelCampo, line 16: Julian and I said a 4 would be better and we set the error message to "ALREADY_REVOKED_UNLOCK_CODE". As far as I know the 4 hasn't been there for a long time

erica/erica_legacy/pyeric/eric_errors.py
```diff
@@ -420,6 +421,11 @@ class EricAlreadyRevokedError(EricTransferError):
     """
     ERROR_CODE = 11
 
+    # Overwrite initaliser to set custom res_code
+    def __init__(self, res_code=None, eric_response=None, server_response=None, server_err_msg=None):
```
VictorDelCampo, line 425: Parameter res_code not needed right? It will always be the 8 (in your case) or the 4 (in our case)

VictorDelCampo (approved):
This is exactly what Julian and I realized last week (since in the rente-webapp we were not able to map this error when using the erica v2).

This change is part of my PR for connecting the erica v2 to the rente-webapp but I am still working on it because we are also introducing the waiting moment in the frontend.

So yes this is a good change and you can merge it as you implemented and I will then merge into my local branch before committing

PR #178, opened 2022-07-15
Log error code from elster

JulianRoesner:
# Short Description
We want to log error codes from the server response from Elster. 

# Changes
- Add a function to extract the codes from the Error xml.
- Log the extracted codes

# Feedback
- Did I miss something important?

### DIFF
--- a/tests/erica_legacy/pyeric/test_eric_errors.py
+++ b/tests/erica_legacy/pyeric/test_eric_errors.py
@@ -3,7 +3,8 @@ import unittest
 from erica.erica_legacy.pyeric.eric_errors import EricGlobalError, EricProcessNotSuccessful, EricGlobalValidationError, \
     EricGlobalInitialisationError, EricTransferError, EricCryptError, EricIOError, EricPrintError, \
     EricNullReturnedError, EricAlreadyRequestedError, EricAntragNotFoundError, check_result, EricUnknownError, \
-    check_xml, EricInvalidXmlReturnedError, EricAlreadyRevokedError, check_handle, EricWrongTaxNumberError
+    check_xml, EricInvalidXmlReturnedError, EricAlreadyRevokedError, check_handle, EricWrongTaxNumberError, \
+    get_error_codes_from_server_err_msg
 from tests.utils import read_text_from_sample
 
 _VALIDATION_ERROR_CODE = 610001002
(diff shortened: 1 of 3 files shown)


tests/erica_legacy/pyeric/test_eric_errors.py
```diff
@@ -276,3 +277,22 @@ def test_if_empty_xml_then_raise_no_error(self):
...
+        result = get_error_codes_from_server_err_msg(server_err_msg)
+        assert result == []
+
+    def test_if_one_error_code_in_server_err_msg_then_return_code_in_array(self):
+        server_err_msg = '<?xml version="1.0" encoding="UTF-8"?><EricGetErrormessagesFromXMLAnswer xmlns="http://www.elster.de/EricXML/1.0/EricGetErrormessagesFromXMLAnswer">\t<Fehler>\t\t<Code>01</Code>\t\t<Meldung>Ein Fehler ist aufgetreten</Meldung>\t</Fehler>   </EricGetErrormessagesFromXMLAnswer>'
+        result = get_error_codes_from_server_err_msg(server_err_msg)
+        assert result == ['01']
+
+    def test_if_many_error_code_in_server_err_msg_then_return_all_codes_in_array(self):
+        server_err_msg = '<?xml version="1.0" encoding="UTF-8"?><EricGetErrormessagesFromXMLAnswer xmlns="http://www.elster.de/EricXML/1.0/EricGetErrormessagesFromXMLAnswer">\t<Fehler>\t\t<Code>01</Code>\t\t<Meldung>Ein Fehler ist aufgetreten</Meldung>\t</Fehler>   <Fehler>\t\t<Code>02</Code>\t\t<Meldung>Ein weiterer Fehler ist aufgetreten</Meldung>\t</Fehler></EricGetErrormessagesFromXMLAnswer>'
```
frederike-ramin, line 295: Are those real Elster XMLs that are possibly returned for a transfer error? I think they're rather self-made, no? I feel like it would be safer to actually have a possible response in our /samples folder and actually pass that to the function - maybe just as an additional test.

frederike-ramin:
As this is not a major issue and we're again getting some errors where this log would be helpful, I'm just gonna merge this, we can still open another PR with the sample data if we want to.

tests/erica_legacy/pyeric/test_eric_errors.py
JulianRoesner: This is a sample xml. I just changed the error codes to make it more readable

frederike-ramin: I see. Great that this is already merged then. 👍  :D

PR #185, opened 2022-07-19
Add optional tax_year to unlock_code_request

frederike-ramin:
# Short Description
- Users of grundsteuer are confused because in their FSC letter it says "Veranlagungsjahr: 2021", although they ant to do the declaration for 2022 - to be able to configure that, I introduced an optional input parameter to differ from "2021" - this way it should work as normal until you provide a different tax_year.

# Changes
- Add tax_year to all unlock_code_request input tapes
- Set tax_year in unlock_code_request XML

# Feedback
- I'm slightly confused if I introduced the default value in the right places. That I have to alter a type a full *three* times seems quite a lot to me and I' not sure I did this correctly.
- Do you see a good place to test that the input tax_year in the end results in the XML and that the default is set correctly? I'm not really satisfied with the test coverage on this one.

### DIFF
--- a/erica/application/FreischaltCode/FreischaltCode.py
+++ b/erica/application/FreischaltCode/FreischaltCode.py
@@ -8,12 +8,13 @@ from erica.application.base_dto import BaseDto
 
 # Input
 from erica.domain.Shared.tax_id_number import TaxIdNumber
+from erica.erica_legacy.elster_xml.elster_xml_generator import VERANLAGUNGSJAHR
 
 
 class FreischaltCodeRequestPayloadDto(BaseDto):
     tax_id_number: TaxIdNumber
     date_of_birth: date
-    tax_year: str = "2021"
+    tax_year: str = str(VERANLAGUNGSJAHR)
 
 
 class FreischaltCodeActivatePayloadDto(BaseDto):
--- a/erica/application/FreischaltCode/FreischaltCode.py
+++ b/erica/application/FreischaltCode/FreischaltCode.py
@@ -8,13 +8,12 @@ from erica.application.base_dto import BaseDto
 
 # Input
 from erica.domain.Shared.tax_id_number import TaxIdNumber
-from erica.erica_legacy.elster_xml.elster_xml_generator import VERANLAGUNGSJAHR
 
 
 class FreischaltCodeRequestPayloadDto(BaseDto):
     tax_id_number: TaxIdNumber
     date_of_birth: date
-    tax_year: str = str(VERANLAGUNGSJAHR)
+    tax_year: Optional[str]
 
 
 class FreischaltCodeActivatePayloadDto(BaseDto):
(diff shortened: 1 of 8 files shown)


tests/erica_legacy/elster_xml/test_elster_xml_generator.py
```diff
@@ -984,7 +985,7 @@ def test_add_vast_antrag_nutzdaten(self):
         xml_string = _pretty(xml_top)
 
         self.assertEqual('false', xml_top.find('.//SpezRechtAntrag/Veranlagungszeitraum/Unbeschraenkt').text)
-        self.assertEqual(str(TEST_EST_VERANLAGUNGSJAHR), xml_top.find('.//SpezRechtAntrag/Veranlagungszeitraum/Veranlagungsjahre/Jahr').text)
+        self.assertEqual(str(self.valid_user_data['tax_year']), xml_top.find('.//SpezRechtAntrag/Veranlagungszeitraum/Veranlagungsjahre/Jahr').text)
```
JulianRoesner, line 988: Isn't that the test you are looking for?

erica/application/FreischaltCode/FreischaltCode.py
```diff
@@ -13,6 +13,7 @@
 class FreischaltCodeRequestPayloadDto(BaseDto):
     tax_id_number: TaxIdNumber
     date_of_birth: date
+    tax_year: str = "2021"
```
JulianRoesner, line 16: Why not use the constant VERANLAGUNGSZEITRAUM for all these places?

erica/erica_legacy/elster_xml/elster_xml_generator.py
```diff
@@ -274,7 +274,7 @@ def _add_vast_request_xml_nutzdaten(xml_top, user_data):
     veranlagungszeitraum_xml = SubElement(spez_recht_antrag_xml, 'Veranlagungszeitraum')
     SubElement(veranlagungszeitraum_xml, 'Unbeschraenkt').text = 'false'
     veranlagungsjahre_xml = SubElement(veranlagungszeitraum_xml, 'Veranlagungsjahre')
-    SubElement(veranlagungsjahre_xml, 'Jahr').text = str(VERANLAGUNGSJAHR)
+    SubElement(veranlagungsjahre_xml, 'Jahr').text = user_data['tax_year']
```
JulianRoesner, line 277: Couldn't we also keep the tax_year optional and the only use it here if it is set? `user_data.get('tax_year', str(VERANLAGUNGSJAHR))`?

JulianRoesner (approved):
Fine with me and great idea. I do not see many other options than having a constant or setting an optional.
Regarding the tests, we also do not test that for other inputs. The only additional thing that you could test is that the correct tax_year is set in user_data depending on the input.

tests/erica_legacy/elster_xml/test_elster_xml_generator.py
frederike-ramin: Kind of - without the API part. But close enough. :)

erica/application/FreischaltCode/FreischaltCode.py
frederike-ramin: Honestly, because Honestly - because I don't fully understand layered Architecture and thought this might be a violation of the one-direction-access rule? But maybe that's Quatsch. I'll use it.

PR #196, opened 2022-08-03
Refactor erica structure

JulianRoesner:
# Short Description
We have updated the structure of the erica. We are proposing the following structure of the erica module:
- Separate folders for everything that is needed by the erica_api (taking post and get request and saving, reading to/from database) and erica_worker (actually sending information to Elster, updating the database)
- Shared folder for everything that both api and worker need (e.g. the database, huey, the ericaRequest model)
- Remove everything from the erica_legacy folder. It is not deleted yet, because it still has everything related to v1. This should be deleted when we no longer need to provide v1 services
- Rename all files/folders to snake_case

@SannyNguyenHung, @nfelger Just have a look if you are interested

Open tasks:
- [ ] Merge the corresponding infra PRs (s. [here](https://github.com/digitalservicebund/erica-infra/pull/6) and [here](https://github.com/digitalservicebund/steuerlotse-infra/pull/64))
- [x] Delete erica_legacy code

# Changes
- Adapt structure and tests
- Remove some files that were not longer needed

# Feedback
- What do you think about the structure in general?
- We are not sure about the naming (erica/erica_shared or erica/erica_api/api)? Do you think we should adapt it more (e.g. have just api, shared and worker as folder names and then put everything in erica_api/api one folder above)?


JulianRoesner:
> Actually, the double erica always confused me. So I'd vote for erica/api 

And do you have an opinion on what we should do with the corresponding erica_api/api folder? Just leave it there?

> Otherwise just one tiny hint: Don't forget to adapt the gitignore to ignore the eric libs. :)

You mean when changing the folder names again? Or did we already miss a second gitignore file?

frederike-ramin:
> And do you have an opinion on what we should do with the corresponding erica_api/api folder? Just leave it there?

Either move it up or find a more fitting name why this even is a specific folder. In the erica/api folder I'd expect folders like /endpoints /dto and /responses on one level. So yes, pull it up. Not sure if we can wrap the single files in there in another folder though. `exception_handling` and error can even go togehter, I think? `utils` makes sense on the top layer, so yeah TLDR: move everything in there one level up, I'd say.

> You mean when changing the folder names again? Or did we already miss a second gitignore file?

Never mind me, I just forgot to move the lib files 🤦‍♀️

JulianRoesner:
@VictorDelCampo I have changed the folder structure to the way we discussed yesterday and Frede proposed above. Wdyt?

Furthermore, I deleted a really old gitignore file. Do you see any problem with that (@frederike-ramin @VictorDelCampo )?

nfelger:
This is a great improvement and simplification! More can always be done, but this is a great step in the right direction, and it feels like this is big enough of a step at this point.

> What do you think about the structure in general?

Without any judgement, just making sure I understand it:
- traditional layers are gone
- instead, we group by deployed unit: api, worker, library code shared by both

I think that's very appropriate for this project.

Seeing it simplified this far, I have some additional thoughts:

1) Is it just me or does the erica_shared package feel like the domain model (incl. persistence infrastructure)? How would you feel about calling this "domain"? I have mixed feelings about the name "shared" because - like e.g. "utils" - it's an invitation to dump anything in here that doesn't obviously belong anywhere else.

2) How would you feel about pulling erica/cron.d up by one level? The details of the cron jobs seem more of are a hosting concern, not an application concern, and could live outside the python package.

3) If I understand the thought behind this structure correctly, we should only have dependencies `api -> shared` or `worker -> shared`. There should be no dependencies between `api` and `worker` and ideally no dependencies from `shared` to either of the others. Do you agree?

I found several of these "illegal" dependencies.

`worker -> api`
Screenshot 2022-08-04 at 14 41 28

`api -> worker`
Screenshot 2022-08-04 at 14 45 37

`shared -> api`
Screenshot 2022-08-04 at 14 41 59

`shared -> worker`
Screenshot 2022-08-04 at 14 45 03

What are your thoughts on these? 

> We are not sure about the naming (erica/erica_shared or erica/erica_api/api)? Do you think we should adapt it more (e.g. have just api, shared and worker as folder names and then put everything in erica_api/api one folder above)?

I agree with Frede, there's too much erica in erica. :) The redundancy makes it hard to read. I also agree with pulling the contents of erica_api/api up one level.

JulianRoesner:
>Seeing it simplified this far, I have some additional thoughts:
>
>1. Is it just me or does the erica_shared package feel like the domain model (incl. persistence infrastructure)? How would you feel about calling this "domain"? I have mixed feelings about the name "shared" because - like e.g. "utils" - it's an invitation to dump anything in here that doesn't obviously belong anywhere else.
>
>2. How would you feel about pulling erica/cron.d up by one level? The details of the cron jobs seem more of are a hosting concern, not an application concern, and could live outside the python package.
>
>3. If I understand the thought behind this structure correctly, we should only have dependencies api -> shared or worker -> shared. There should be no dependencies between api and worker and ideally no dependencies from shared to either of the others. Do you agree?

thank you for your input. I will try to answer those in a different order :) :
2. Yes, we could do that. We also discussed both possibilities and I would say had no strong argument against moving it up besides just keeping it the old way because this did not produce that many problems. So, yay.
3. Yes, that was the idea and would be the best case. However, the problem is that there is still too much coupled. For example the jobs function itself (what is happening when a job is performed) need to be called when the api is getting a request. We thought that it be better to still keep that in the worker to make clear that this is where the function is happening and where the responsibility is (it is basically the export function of the worker 🤔) and then call it from the outside. To reduce all of those dependencies would probably mean a lot of refactoring (e.g. splitting up the JobService into two functions) that we thought might be a bit too much for now.
That said, I will have another look into the imports and I will check whether we could reduce them. What do you think?

1. If we would put more into the shared folder to reduce the dependencies this would be more than just the domain. I think it already is right now. However, we could keep repositories, payloads, model in a folder domain and just move the job_service in its own folder under /erica. This would also remove the very general naming of the shared folder which I also do not like that much. Wdyt about that?

nfelger:
> 3. Yes, that was the idea and would be the best case. However, the problem is that there is still too much coupled. For example the jobs function itself (what is happening when a job is performed) need to be called when the api is getting a request. We thought that it be better to still keep that in the worker to make clear that this is where the function is happening and where the responsibility is (it is basically the export function of the worker 🤔) and then call it from the outside. To reduce all of those dependencies would probably mean a lot of refactoring (e.g. splitting up the JobService into two functions) that we thought might be a bit too much for now.
> That said, I will have another look into the imports and I will check whether we could reduce them. What do you think?

I don't fully understand the details, but given that we don't want to invest too much more time in this, it seem like we have two options:
1) accept that the package boundaries are leaky and imperfect,
2) do the smallest amount to work to create hard boundaries, which probably means making the "shared" package more of a dumping ground.

Would you agree with this? If so, which course of action do you think will make it easier for someone to understand erica's structure at some point in the future (which I believe should be our goal)?

> If we would put more into the shared folder to reduce the dependencies this would be more than just the domain. I think it already is right now. However, we could keep repositories, payloads, model in a folder domain and just move the job_service in its own folder under /erica. This would also remove the very general naming of the shared folder which I also do not like that much. Wdyt about that?

Sounds sensible to me and would keep the door open to a more considered move back towards a layered/clean architecture at some time in the future!

JulianRoesner:
> I don't fully understand the details, but given that we don't want to invest too much more time in this, it seem like we have two options:
>
> 1. accept that the package boundaries are leaky and imperfect,
> 2. do the smallest amount to work to create hard boundaries, which probably means making the "shared" package more of a dumping ground.

Yes, I was not very clear in my explanation. Personally, I think we should go with one because imo it is easier to understand that there are these two components and grasp what they do if they contain most of their code instead of a overloaded shared folder.
WDYT, @frederike-ramin @VictorDelCampo ?

VictorDelCampo:
> > I don't fully understand the details, but given that we don't want to invest too much more time in this, it seem like we have two options:
> > 
> > 1. accept that the package boundaries are leaky and imperfect,
> > 2. do the smallest amount to work to create hard boundaries, which probably means making the "shared" package more of a dumping ground.
> 
> Yes, I was not very clear in my explanation. Personally, I think we should go with one because imo it is easier to understand that there are these two components and grasp what they do if they contain most of their code instead of a overloaded shared folder. WDYT, @frederike-ramin @VictorDelCampo ?

It depends on how much time do we have for erica right now... for me personally would be easier to understand having only the "api", the "worker" and the "domain" (shared parts). Now we also have:

- cron.d (I will move it one layer up)
- legacy (I will delete it soon since v1 is no longer in use)
- job_service

This last one, as Julian explained, contains function used in both components ("api" and "worker"). We thought about splitting it up but this would mean more than a minor refactor (which was not the objective of this task). But on the other hand I think we should leave erica as understandable as possible before (at least) the rente team stops working on rente (end of August).

I will talk to Sanny to determine the open tasks in rente for the next 3 weeks and if I have enough time I will definitely split up the job_service.

WDYT @frederike-ramin ?

VictorDelCampo:
- [x] Cron.d moved 1 layer up (also updated Dockerfile)
- [x] Removed legacy code including tests
- [x] Removed unused gitignores

frederike-ramin:
As I have been asked "WDYT" a lot of times now, this is WIT:
- This improves Erica greatly as is.
- I personally would not put in the effort to further separate the job_service. In my head, it was even a bit more understanding for newcomers with only three folders (and cron, which makes sense) but I see the reasons for an extra job_service folder.
- Honestly, I think this is the point where we should stop. We know that there are dependencies between folders, but at least you might know where to start when you get to work on Erica.
- That being said, @VictorDelCampo you seem passionate about splitting up the job_service folder and I won't stop you, but I don't think we should do it in this PR^^

TLDR: I think: Merge as is. ;)

VictorDelCampo:
@frederike-ramin Agree with you about merging the current state before do any greater refactor because of the job_service

PR #230, opened 2022-09-06
Allow access to eric_log_path from verify_using_stick

JulianRoesner:
# Short Description
We want to log what has happened when and only when the verify dongle does not work.

# Changes
- Store temp log path in Eric Wrapper
- Log eric logs when error occurs during verify stick

# Feedback
- Do you see any problem with that?

### DIFF
--- a/erica/worker/pyeric/eric.py
+++ b/erica/worker/pyeric/eric.py
@@ -41,7 +41,7 @@ class EricVerschluesselungsParameterT(Structure):
 
 # TODO: Unify usage of EricWrapper; rethink having eric_wrapper as a parameter
 @contextmanager
-def get_eric_wrapper(keep_logs=False):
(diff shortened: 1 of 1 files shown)


frederike-ramin:
Could you give me a little more context what you're trying to achieve with that? Simply gather more info about why the stick is not available? If so, I don't see a problem with that. The only thing I could think of is logging user-specific data, but we don't have any with verify_using_dongle, so 👍

JulianRoesner:
> Could you give me a little more context what you're trying to achieve with that? Simply gather more info about why the stick is not available? If so, I don't see a problem with that. The only thing I could think of is logging user-specific data, but we don't have any with verify_using_dongle, so 👍 

Exactly. The hope is that we might find some information in the Eric log

PR #245, opened 2022-11-03
Make SteuerID in grundsteuer input optional.

shrzaf:
The requirements are specified in https://steuerlotse.atlassian.net/browse/STL-2902?focusedCommentId=13022 

To recap:

SteuerID given: test/real submissions are determined using the 0 prefix (same as before).

SteuerID not given: environment-specific config determines whether to mark the submission as test (false on production, true on staging)


JulianRoesner (approved):
LGTM 🚀 I like the added tests 👍

PR #246, opened 2022-11-07
Fix person validation on missing SteuerID.

shrzaf:
# Short Description
- Eigentümer with no SteuerID should now pass validation.
- One test case, `test_if_all_persons_have_test_tax_id_number_not_special_testmerker_then_raise_error`, did not seem to be doing what it claimed to and was equivalent to another test case further down, so I appropriated it to define a new test case.

### DIFF
--- a/tests/api/dto/test_grundsteuer_input_eigentuemer.py
+++ b/tests/api/dto/test_grundsteuer_input_eigentuemer.py
@@ -100,11 +100,11 @@ class TestEigentuemer:
                           steuer_id="03352417692", anteil=Anteil(zaehler="1", nenner="2"))]
         Eigentuemer.parse_obj({"person": persons})
 
-    def test_if_all_persons_have_test_tax_id_number_not_special_testmerker_then_raise_error(self):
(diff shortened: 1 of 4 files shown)


tests/api/dto/test_grundsteuer_input_eigentuemer.py
```diff
@@ -100,11 +100,11 @@ def test_if_all_persons_have_test_tax_id_number_then_raise_no_error(self):
                           steuer_id="03352417692", anteil=Anteil(zaehler="1", nenner="2"))]
         Eigentuemer.parse_obj({"person": persons})
 
-    def test_if_all_persons_have_test_tax_id_number_not_special_testmerker_then_raise_error(self):
+    def test_if_all_persons_have_test_tax_id_number_or_none_then_raise_no_error(self):
```
JulianRoesner, line 103: Yes, I am fine with that. I think the old test was some legacy test that we probably should have adapted earlier on. Thanks

JulianRoesner (approved):
Nice. LGTM 🚀

PR #277, opened 2022-12-14
Install pcsc from debian testing (bookworm).

shrzaf:
# Short Description
Upgrades [pcscd](https://packages.debian.org/bookworm/pcscd) and [pcsd-tools](https://packages.debian.org/bookworm/pcsc-tools) from bullseye (current debian stable) to bookworm (current debian testing). This is an attempt to fix the underlying issue that triggers the mysterious ERIC_CRYPT_E_SC_SLOT_EMPTY error.

# Changes
- Adds the bookworm repository to apt sources.
- Installs pcscd and pcsc-tools from bookworm. Apart from these two and their dependencies (see links above for the list), sources of the rest of the packages should remain unchanged (i.e. bullseye).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ ENV ELSTER_DATENLIEFERANT=$elster_datenlieferant
 ENV ELSTER_HERSTELLER_ID=$elster_hersteller_id
 
 WORKDIR /app
-RUN echo "APT::Default-Release \"bullseye\";" >> /etc/apt/apt.conf.d/default-release && \
+RUN echo "Package: *\nPin: release n=bookworm\nPin-Priority: 50\n" >> /etc/apt/preferences && \
     echo "deb http://ftp.debian.org/debian bookworm main" >> /etc/apt/sources.list.d/bookworm.list && \
     apt-get update && \
     apt-get install --no-install-recommends -y -t bookworm pcsc-tools pcscd && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,11 @@ ENV ELSTER_DATENLIEFERANT=$elster_datenlieferant
 ENV ELSTER_HERSTELLER_ID=$elster_hersteller_id
 
 WORKDIR /app
-RUN apt-get update && \
-    apt-get install --no-install-recommends -y pcsc-tools pcscd procps unzip curl && \
+RUN echo "APT::Default-Release \"bullseye\";" >> /etc/apt/apt.conf.d/default-release && \
+    echo "deb http://ftp.debian.org/debian bookworm main" >> /etc/apt/sources.list.d/bookworm.list && \
+    apt-get update && \
+    apt-get install --no-install-recommends -y -t bookworm pcsc-tools pcscd && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -14,8 +14,11 @@ ENV ELSTER_DATENLIEFERANT=$elster_datenlieferant
 ENV ELSTER_HERSTELLER_ID=$elster_hersteller_id
 
 WORKDIR /app
-RUN apt-get update && \
-    apt-get install --no-install-recommends -y pcsc-tools pcscd procps unzip curl && \
+RUN echo "APT::Default-Release \"bullseye\";" >> /etc/apt/apt.conf.d/default-release && \
+    echo "deb http://ftp.debian.org/debian bookworm main" >> /etc/apt/sources.list.d/bookworm.list && \
```
JulianRoesner, line 18: This adds the bookworm repository to allow us using it in line 20 to install the pcsc-tools etc?

JulianRoesner (approved):
LGTM as far as I can see

Dockerfile
shrzaf: This sets bullseye as the default repository so that running `apt-get install `without the -t options continues using bullseye as the default source. You could verify this by building the image (easiest would be to build the first stage only) with and without this change and comparing versions of packages procps, unzip and curl (installed in line 21) inside the resulting container. [In](https://packages.debian.org/bullseye/curl) [all](https://packages.debian.org/bullseye/unzip) [cases](https://packages.debian.org/bullseye/procps) the bullseye version should have been installed.

JulianRoesner: Ah, thanks. The naming of bullseye and bookworm is too close and I mixed them up. That makes sense. Tbh I have not interacted with the apt configuration so far. Is there a reason that we are not running into the problem described [here ](https://unix.stackexchange.com/a/647205) and should use pinning because of that?

shrzaf: Good point. I think we're not really affected by this since the debian installation is only going to run in a (relatively) short-living container where no apt upgrade is ever executed or needed, but pinning would still be the better approach and it's easy enough to implement. Commit will come in a bit!

Dockerfile
```diff
@@ -14,7 +14,7 @@ ENV ELSTER_DATENLIEFERANT=$elster_datenlieferant
 ENV ELSTER_HERSTELLER_ID=$elster_hersteller_id
 
 WORKDIR /app
-RUN echo "APT::Default-Release \"bullseye\";" >> /etc/apt/apt.conf.d/default-release && \
+RUN echo "Package: *\nPin: release n=bookworm\nPin-Priority: 50\n" >> /etc/apt/preferences && \
```
JulianRoesner, line 17: I would have assumed that we would raise the pin priority of bullseye but we are setting the priority of bookworm quite low, correct? 


LGTM

shrzaf: This is the solution proposed in the SO thread you mentioned. We're assigning a lower priority to bookworm so that apt would never prefer it to the bullseye sources unless explicitly instructed to.

PR #284, opened 2023-01-03
Make invalid aktenzeichen a special error code

frederike-ramin:
# Short Description
- Currently, we're still outputting a not-very-readable error message if the Aktenzeichen is wrong. (see [this issue](https://github.com/digitalservicebund/grundsteuer/issues/685)) We intend to change that in Grundsteuer, but I thought it'd be good to already catch the corresponding error in Erica.

# Changes
- Add a new error code explicitly for invalid Aktenzeichen + test

# Feedback
- I'm not 100% sure I like that once we have this specific validation error, we're loosing possible other ones. We're not aware of other errors as far as I know, but once we have an unknown validation error *plus* this one, users will have to retry to get the remaining errors. However, I didn't like the idea of nesting this even further into the `generate_error_response` of the `EricGlobalValidationError` - wdyt?

### DIFF
--- a/erica/worker/pyeric/eric_errors.py
+++ b/erica/worker/pyeric/eric_errors.py
@@ -14,6 +14,7 @@ _ERIC_CUSTOM_ERROR_CODES = {
     5: "ELSTER_REQUEST_ID_UNKNOWN",
     6: "INVALID_BUFA_NUMBER",
     7: "INVALID_TAX_NUMBER",
+    8: "INVALID_AKTENZEICHEN",
 }
 
 _ERIC_GLOBAL_VALIDATION_ERRORS = {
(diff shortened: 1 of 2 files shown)


frederike-ramin:
(Feel free to ignore the failing tests, I'm changing the certificate, locally, they worked)

JulianRoesner (commented):
> I'm not 100% sure I like that once we have this specific validation error, we're loosing possible other ones. We're not aware of other errors as far as I know, but once we have an unknown validation error plus this one, users will have to retry to get the remaining errors. However, I didn't like the idea of nesting this even further into the generate_error_response of the EricGlobalValidationError - wdyt

I think I do not really understand why we would lose other errors bc according to the documentation this is exactly the error code for that problem. But I might be missing something.


I also do not think that nesting it makes much sense from an outside perspective because we want to differentiate between general validations errors ( -> showing users the actual error message) and precisely defined error messages where we know exactly what is happening. Therefore, I think that is fine. Internally we might structure them together but I would rather not add more inheritance here.

The one thing I am wondering is whether we should add the handling of the error to grundsteuer first.

frederike-ramin:
We decided to let Grundsteuer handle this directly. As checking the Aktenzeichen is not a specific Erica task (yet) we wouldn't expect the validation to return such a specific error (in contrast to the Steuernummer error). Done there: [Improve Aktenzeichen error message](https://github.com/digitalservicebund/grundsteuer/commit/70c11e24cf5e450a0c2088f87e543788b1f72413)

